PR #1, opened 2014-10-29
Use tokenized PHP in Customizer settings checks rather than regex.

kovshenin:
As noted in the theme reviewers mailing list the regex in the customizer check easily chokes when a semicolon is encountered anywhere before the end of the method call (comment, default value, etc.).

This is an attempt at using tokenized PHP to validate sanitize_callback or sanitize_js_callback exist and are not empty.

### DIFF
--- a/checks/customizer.php
+++ b/checks/customizer.php
@@ -9,34 +9,90 @@ class CustomizerCheck implements themecheck {
 
 	function check( $php_files, $css_files, $other_files) {
 
-		$ret = true;
-
 		checkcount();
 
 		/**
 		 * Check whether every Customizer setting has a sanitization callback set.
 		 */
 		foreach ( $php_files as $file_path => $file_content ) {
+
 			// Get the arguments passed to the add_setting method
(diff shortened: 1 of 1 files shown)


Otto42:
Given how badly the i18n checks are performing (which also use the tokenizer), I'm now much less inclined to trust it. Thinking we need an overall better solution here, or to simplify these checks to be less precise somehow.

Otto42:
The new textdomain code uses the tokenizer and doesn't seem to be breaking things. Perhaps we can integrate these checks in there and kill two birds with one stone.

Otto42:
Related #87. Given that I'm also using the tokenizer in the text domain checks now, a generic approach might be better to avoid duplicated work. Still looking at it. Maybe pass tokenizer data around? Still, discuss in #87.

Otto42:
Closing this in favor of a future redesign as a whole, which uses the tokenizer more extensively and avoids having to retokenize the files over and again. Also will reduce memory usage and simplify some of the checks.

PR #2, opened 2014-10-29
Add WP-CLI support.

wokamoto:
## SYNOPSIS

```
wp theme-check [<theme>] [--format=<format>] [--output=<output>]
```
## OPTIONS

```
[<theme>]
 Theme name. Default: current theme

[--format=<format>]
 Accepted values: text, json. Default: text
```
## EXAMPLES

```
wp theme-check twentyfourteen
```

### DIFF
--- a/wp-cli.php
+++ b/wp-cli.php
@@ -17,9 +17,6 @@ class Theme_Check_Commands extends WP_CLI_Command {
 	 * [--format=<format>]
 	 * : Accepted values: text, json. Default: text
 	 *
-	 * [--output=<output>]
-	 * : Accepted values: stdout, stderr. Default: stdout
-	 *
 	 * ## EXAMPLES
 	 *
 	 * wp theme-check twentyfourteen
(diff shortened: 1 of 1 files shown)


Otto42:
Is this still relevant or okay given the current state? I'm fine with adding the needed hooks, but given that we now use the unaltered plugin on org, I'm hesitant to merge without detailed analysis. Especially given the large number of changes here.

wp-cli.php
```diff
@@ -0,0 +1,129 @@
...
+	 * : Theme name. Default: current theme
+	 *
+	 * [--format=<format>]
+	 * : Accepted values: text, json. Default: text
+	 *
+	 * ## EXAMPLES
+	 *
+	 * wp theme-check twentyfourteen
+	 */
+	function __invoke($args, $assoc_args) {
```
GaryJones: This method is far too big and could be broken down into many smaller methods.

Issue #4, opened 2014-10-31
Don't check "node_modules" folder


devinsays:
A lot of developers use Grunt for automating tasks. It would be nice to warn that "node_modules" folder needs to be removed, but skip any checks inside since developers are likely just using this in development (and it takes a while if you forget to remove it).

PR #9, opened 2014-11-10
Add node_modules to filename checks

ryelle:
This also prevents scanning into `node_modules` and `.sass-cache` folders (but keeps the top-level folder for the filename check). See #4 for background.

### DIFF
--- a/checks/filenames.php
+++ b/checks/filenames.php
@@ -18,19 +18,20 @@ class File_Checks implements themecheck {
 			array_push( $filenames, strtolower( basename( $php_key ) ) );
 		}
 		$blacklist = array(
-				'thumbs.db'				=> __( 'Windows thumbnail store', 'theme-check' ),
-				'desktop.ini'			=> __( 'windows system file', 'theme-check' ),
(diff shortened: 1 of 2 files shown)


Otto42:
If you want to submit a patch to ignore files on a plugin only manner, I'll accept that, after feedback. This is too strong. It can't be in the checks themselves. The checks are identical on w.org and I want to keep them that way. Put it in the base files, not in the checks.

Otto42:
I didn't mean to close that. Just need a more targeted patch.

retlehs:
@devinsays good idea. `bower_components` is another one that should be ignored, and is usually also found in a `.gitignore`

Otto42:
No traction on this. If this is useful, let's define a generic approach, or a plugin specific thing that can read various ignore files.

grantpalin:
Also consider the case where packages are located elsewhere than the default - Bower can do this, not sure of NPM can also. A per-project file seems the best way to manage the ignores.

twobyte:
Keeping front-end source files outside of theme folder but inside `wp-content/themes` throws a broken theme / no stylesheet error in Wordpress. Do we have a way to specify theme folders WP should ignore?

Issue #3, opened 2014-10-31
Unmatched parentheses in i18n check break the regex


Otto42:
Reported here:
https://wordpress.org/support/topic/bug-i-guess?replies=2#post-6175594

Example:
_e( 'text field (you can leave it blank', 'mytheme' );

Likely solution: tokenize the whole file instead of just that line. Similar to the patch for this:
https://github.com/Otto42/theme-check/pull/1

Otto42:
Given the speed problems this already has, no. Let it break. Maybe detect it and fail themes that are doing it, but strings in translation need to be complete, not partial. Wontfix.

PR #12, opened 2014-11-17
Updates to textdomain checks

ryelle:
This pull request uses a different regular expression for finding the translation functions and strings, and combines the I18NCheck and TextDomainCheck, with some added features.

* This checks `__()`, `_e()`, `_x()`, `_n()`, and `_nx()` for textdomains
* It looks for and checks against the theme's Text Domain (as defined in style.css) or the theme slug, throwing a warning (recommended) if the textdomain doesn't match. This may be overly aggressive, since theme devs can have a second textdomain for frameworks, and this would alert for each use of that allowed extra. In my use, though, it's been helpful to locate extra invalid domains.
* Using the tokenizer from I18NCheck, make sure the textdomain is not a variable, constant, or unquoted string.
* Finally, count and list all unique textdomains used by the theme, and raise a warning (recommended) if there are more than 2.

This doesn't get stuck on parentheses like mentioned in #3, but like the existing regular expression it does skip strings with quotes-- ` _e( "I'm an example" );` is not detected.


Otto42:
The text domain checks have been completely rewritten using the tokenizer now, so regex is not needed for them any longer. The i18n checks still need a reworking, but will likely be reworked in the same manner.

PR #33, opened 2014-12-19
changes for consistent design

kniebremser:


Otto42:
All of these need to have issues associated with them and they need to be consolidated into a single pull request, with explanations and such. I don't accept "random" pull requests on this repository.

PR #37, opened 2015-01-08
Change Plugin URI

utsumit:
I noticed Plugin URI is dead.
http://ottopress.com/wordpress-plugins/theme-check/

So I changed the URI to the plugin page.
https://wordpress.org/plugins/theme-check/

Changing the URI to the github page is also good.
https://github.com/Otto42/theme-check

### DIFF
--- a/lang/theme-check.pot
+++ b/lang/theme-check.pot
@@ -929,7 +929,7 @@ msgid "Theme Check"
 msgstr ""
 
 #. Plugin URI of the plugin/theme
-msgid "http://ottopress.com/wordpress-plugins/theme-check/"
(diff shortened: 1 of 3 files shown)


Otto42:
The "Plugin URI" is not the same as a "Plugin URL". It is an identifier, it does not need to point to a live webpage (although that certainly is usually the case).

This is not necessary, and in fact undesirable. URIs should be unique and unchanging identifiers, and also should not really be pointing to URLs on WordPress.org.

So, I'm rejecting this pull request.

utsumit:
Sorry, I misunderstood "Plugin URI".

PR #41, opened 2015-01-18
Updated style tags check

zauan:
Please see : https://github.com/Otto42/theme-check/issues/39

The global $data already contains information about the current theme checked.


Otto42:
Using the global like this works for the plugin, but not for other cases where the checks are used independently, without the rest of the plugin code. Checks must be self-contained and functional on a standalone basis.

PR #42, opened 2015-01-21
Add a check for `.screen-reader-text`.

obenland:
Six years ago, `.screen-reader-text` was introduced as a canonical class name for text targeted to screen readers for all of core.
Unknown to many, this includes the front-end by virtue of its use in `get_search_form()`.

To ensure forward compatibility with future a11y improvements, and be able to deal with child themes adding theme support for HTML5 features which rely on that class heavily, all themes should have it defined to deal with assistive text.

See https://core.trac.wordpress.org/ticket/29699.

### DIFF
--- a/checks/style_needed.php
+++ b/checks/style_needed.php
@@ -22,6 +22,7 @@ class Style_Needed implements themecheck {
 			'\.wp-caption' => __( '<strong>.wp-caption</strong> css class is needed in your theme css.', 'theme-check' ),
 			'\.wp-caption-text' => __( '<strong>.wp-caption-text</strong> css class is needed in your theme css.', 'theme-check' ),
 			'\.gallery-caption' => __( '<strong>.gallery-caption</strong> css class is needed in your theme css.', 'theme-check' )
+			'\.screen-reader-text' => __( '<strong>.screen-reader-text</strong> css class is needed in your theme css. See See: <a href="http://codex.wordpress.org/CSS#WordPress_Generated_Classes">the Codex</a> for an example implementation.', 'theme-check' )
 		);
 
 		foreach ($checks as $key => $check) {


joedolson:
I'd suggest not just checking that the style is defined, but also that the definition does *not* include either display: none or visibility: hidden, since either of those characteristics would defeat the purpose of the class. Those are the only styles that would be definitely wrong for the class.

Otto42:
Seems a bit strong to me. I'll investigate the core code more, but if this is that important, then should not core have rules for it or auto insert then into the header? What could a theme add to such a rule that would place it in the realm of theming?

joedolson:
I agree that it should be auto-inserted in core; but I haven't been able to get a lot of buy-in for that. https://core.trac.wordpress.org/ticket/29699

Regardless, I think if we're going to check for this in theme-check, we should verify that any styles declared are usable screen-reader-text, since it should be a relatively trivial thing to check for. (Not that I've worked out the regex.)

Otto42:
It would help a lot to know what this CSS should actually be, and what variants of it there might be which would be reasonable.

Looking over that core ticket, I find that the discussion was too much about "add_theme_support" and not enough about the class itself or what the CSS it should contain would be. Is there some universal CSS that it could be? Themes wouldn't need to declare support for it, or include anything, if it was always going to be the exact same CSS code. Because, if that were the case, core could just insert it in the wp_head willy-nilly and problem solved.

joedolson:
Really, there are only two things that it *can't* be - visibility:hidden or display:none. Anything else is totally viable. There's no obligation on behalf of a theme developer that the text wrapped by this class be hidden at all, but if it *is* hidden, it needs to be hidden in a manner that allows screen reader users to continue to access it.

It'll change over time, but right now the recommended code is here: https://make.wordpress.org/accessibility/2015/02/09/hiding-text-for-screen-readers-with-wordpress-core/

Since 'clip' has now been deprecated, we'll be needing to modify it with clip-path soon.

Otto42:
Joe, I'm concerned about backwards compatibility here if themes are required to implement this class. What happens to themes that exist now without it? We're starting to see changes in core that could eventually become breaking changes if this is not addressed.

Reference: 
https://make.wordpress.org/core/2015/03/31/screen-reader-text-in-output-of-comments_popup_link/

Core needs to be adding this class to all pages, in a universal way, for all themes, if we're going to start changing front-end code to use it. I don't see that we can rely on themes to implement this, especially since virtually no theme out there now has it. While comments_popup_link is relatively minor in the grand scheme, if this is going to be added to more elements, then it needs to be addressed and I'm just not seeing the appropriateness of themes doing it.

justintadlock:
By the Theme Review guidelines, themes must define the `.screen-reader-text` class.  Nothing else is required.  The PR looks good and matches our guidelines.

What core does is outside of Theme Review's control, so we need to take steps to make sure theme authors are handling this.

joedolson:
I don't think there's any major risks to this. Theme authors need to learn what this is and how to do it, and it's really extremely easy to do. If they fail to do it, it won't have any significant impact in the vast majority of cases. 

While core wants to promote the use of the class aggressively, it's not so much because we intend to start adding it all over the HTML output; it's because we want to have the option to add it in front-end HTML in the future without needing to worry about backwards compatibility every time. 

Right now, themes that don't define the class will have additional visible text in place where the screen reader text class is used.

obenland:
`screen-reader-text` on the front-end has been around since it was added to the search form in [r11312](https://core.trac.wordpress.org/changeset/11312) six years ago. It's no different than other built-in classes core uses.

I don't think core should output a compat style (which would have to be supported for all eternity), IMO theme author education is the way to go. It is not true that no themes support it. All Twenty themes and there derivatives do, as well as 300,000+ `_s`-based themes, plus many more.

We've been on an education offensive since 4.1 and theme authors have been very receptive.

Otto42:
@obenland: And yet core just broke backward compatibility. Okay, not in a bit way, but they have already done it. Look:


This needs another look. We don't want themes to add support for this. What we should want is for core to just add the .screen-reader-text class inline using wp_head, period. No opt-out.

The purpose of the .screen-reader-text class is to hide content in an accessible manner. This is content that is not supposed to be shown, but which also cannot be display:none'd or visibility:hidden'd. It has to be visible to screen readers, but ideally invisible the rest of the time.

joedolson:
The only way that a theme would legitimately change them would be if they wanted to display the screen reader text for all users. This is entirely legitimate, though an uncommon choice.

I have no problem with auto enqueuing these styles into wp_head from core, although I think that there should be an opt out available via filter. 

But I'm also tired of the argument; lots of people seem to have a problem with adding these enqueued from core, and this argument has been going on for months. I'm all for the theme author education argument, but the end result will be better and more consistent if we enqueue it from core.

obenland:
> Answer me a simple question: In what way, specifically, are those specific style rules for .screen-reader-text in any way theme-specific? How would a theme change them? Under what circumstances?

I can see your point, but where would that be different from `.alignleft`, where the only theme-specific thing might be margins?

Up until 4.2 with emoji's, Core has not added a single script or style to the logged-out output by default, which I think should stay that way. It's a theme's job to style core output, including this. The only time an old theme would not be confronted with that class would be if it used a custom search form and omitted it. In all other cases, it's been around since its introduction six years ago.

That changeset is really not a big deal. Even if a theme does not have styles for screen-reader-text, the worst thing that'll happen is more context for that headline. That could have been a regular enhancement like any other.

obenland:
> the end result will be better and more consistent if we enqueue it from core.

If we'd start doing that, there is no turning back. We will never be able to remove it.
This would make for a very bad developer experience if you have to start removing actions and filter to take ownership of your site's front-end.

Otto42:
> I can see your point, but where would that be different from .alignleft, where the only theme-specific thing might be margins?

Alignleft, right, and center are entirely theme-specific. While it's simple to assume that alignment is simply a matter of fiddling with margins, One might use alignment on things like images to cause them to float, perhaps. Or maybe you're doing something crazy and table-driven. Who knows? Point being that these are clearly layout and within the purview of the theme to decide how best to handle. 

Hiding text for accessibility isn't really a theme thing. It doesn't have anything to do with layout. It doesn't have to do with styling, or color, or font, or anything. It's literally something that is not supposed to be seen on the screen, so it's by definition outside of what the theme should care about.

> If we'd start doing that, there is no turning back. We will never be able to remove it.

Then we should not be adding things like these classes in the first place. Either the front-end is the realm of the theme 100%, or core is allowed to interfere. Core is now adding spans and special classes intended to hide things. Core is already interfering. If we're going to allow that, then it should be done correctly. The theme shouldn't care about things intended to not be seen. Themes shouldn't care about opengraph meta tags. Themes shouldn't care about SEO (much). Themes shouldn't care about the page <title> either, and now they have the option not to do so. 

Well, things that are meant to be hidden on the page to make it more accessible for screen-readers falls into the category of things themes definitely should not care about in my mind.

Otto42:
> What core does is outside of Theme Review's control

Justin, what core does is within everybody's control. This is an open-source project. If we don't speak up when we see things being done wrong, then the result is that they get done wrong.

Putting this on themes makes no sense. Core is doing-it-wrong.

justintadlock:
My point is that it doesn't matter for Theme Check if core is doing it wrong.  If they do it wrong, it's done wrong.  That's a conversation that should be handled on WordPress Trac. 

The purpose of Theme Check is to help theme authors meet the theme review guidelines and to automate some of the things that reviewers would otherwise need to manually check.  The theme review team has had two separate discussions on this requirement and have decided that all themes going forward must have the `.screen-reader-text` class defined.  If/when core changes that, TRT can change it's guidelines.  Until then, there's an existing pull request to the Theme Check plugin.  If this code isn't added to the plugin, reviewers must manually check for the existence of this class, which is kind of crappy for the team.

I personally don't care much if core or themes are left to handle it.  My concern is getting theme authors and theme reviewers on the same page by automating something that is easily automated.  It makes the review process much easier.

Otto42:
I understand that, but I won't do the wrong thing just because it is convenient and easy.

If you want me to draw a line in the sand, then I'll do so: Until I'm convinced that this is the correct direction for WordPress, I will not merge this PR.

justintadlock:
Yes, I'd rather you draw a line in the sand and say that. While TRT doesn't get the automated check, that's actually helpful knowing where you stand on merging the PR.

Otto42:
I really just want to know that this is the right way. I'm not convinced yet, but some talks in #accessibility on slack are encouraging.

Otto42:
As a temporary measure for accessibility, I'm going to go ahead and commit this, however in the long run, core needs to handle these itself. Putting this burden on the themes is the wrong approach. But, we may be able to get a core solution done in the next few releases. https://core.trac.wordpress.org/ticket/29699 is where to monitor.

PR #44, opened 2015-01-22
Check if the file exists before opening it.

obenland:
Can cause a PHP warning, failed to open stream: No such file or directory in themes that reference a file that is not actually there.


Otto42:
How can I reproduce this error? Where is tc_grep or tc_preg getting called where it gets passed a non-existent file?

Otto42:
I'm going to merge this just because it will hide the error, if one exists, but I'd still like to know how we can be calling those functions on non-existent files.

PR #46, opened 2015-02-10
add lack of punctuation marks

kniebremser:
A text string has a dot. The others did not. 
Approximation to a uniform style within the plugin.


Otto42:
Slightly trivial for me, unless it unifies everything in all files for the plugin. I'm not that concerned about style, more about providing information. And, given that translation is now possible, changing strings had to be of substance, rather than trivial punctuation changes.

PR #47, opened 2015-02-12
Translation update nl_NL

jorrit:
Translation update nl_NL. The .mo file still has to be updated.


Otto42:
Translations for the theme check plugin will be handled automatically via the WordPress translate system. 

Contribute your translations at translate.wordpress.org .

PR #53, opened 2015-03-31
Adding extra path reporting

jgraup:
Wrapping filename with file path/link. Adding path reference in a couple cases.


Otto42:
What does this do, exactly? Doesn't seem to add useful info to me, but I might be missing something.

PR #58, opened 2015-05-15
Add check also for `$GLOBALS['content_width']`

grappler:
`_s` uses now `$GLOBALS['content_width']` instead of `$content_width`

https://github.com/Automattic/_s/pull/460

### DIFF
--- a/checks/content-width.php
+++ b/checks/content-width.php
@@ -10,7 +10,7 @@ class ContentWidthCheck implements themecheck {
 		// combine all the php files into one string to make it easier to search
 		$php = implode( ' ', $php_files );
 		checkcount();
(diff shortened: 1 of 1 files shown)


mtomas7:
Thanks! When it will be active on WP.org?

Otto42:
Not until I wrap up the plugin and release a new version.

ScottSmith95:
Is this like something that would happen in a couple days or will it be much longer than that?

Otto42:
I'll try to get a release pushed to w.org sometime this next week.

mtomas7:
Let's hope that WP.org will be updated before the WP4.3.

Otto42:
This specific change was already pushed to w.org. However, it's not in the release version of the plugin yet.

PR #68, opened 2015-06-29
Improve code readability - Patch 1

noplanman:
Make the code easier to read and conform more to the WP coding standards.


Otto42:
No, I don't like most of this change, and I really don't really care for WP's own particular brand of standards. Those are fine if you're writing patches for core, but their standards do not apply to my plugins.

PR #71, opened 2015-07-06
Add the_post_navigation function to post navigation check.

sarahmonster:
As it stands, currently there's no check for `the_post_navigation()` (added in 4.1.0). This means that themes using this function fail the theme check. Since _s is now using this function (https://github.com/Automattic/_s/blob/master/single.php) the check should be updated to allow for this function as well.

### DIFF
--- a/checks/postsnav.php
+++ b/checks/postsnav.php
@@ -14,9 +14,10 @@ class PostPaginationCheck implements themecheck {
 		     strpos( $php, 'paginate_links' ) === false && 
 		     strpos( $php, 'the_posts_pagination' ) === false &&
 		     strpos( $php, 'the_posts_navigation' ) === false &&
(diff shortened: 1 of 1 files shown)


Otto42:
This is incorrect, this check is for posts pagination, as in "older posts" and "newer posts". The single `the_post_navigation()` function shows a link to the next post and previous post, and would be used on the single-post page, for navigating amongst individual posts. You will still need to include something like `the_posts_navigation()` on archive pages, which show more than one post.

If _s is missing a way to navigate pages of posts, then _s is doing-it-wrong.

sarahmonster:
Duh, you're right. :)

Weirdly enough, I was reviewing a theme this morning that gave me this error when I ran it through the plugin—although now it seems to work as expected. I was wondering how they'd managed to upload it without it passing the theme check. 

Chalking this up to "case of Mondays" and closing.

PR #72, opened 2015-07-06
Theme name is often different from theme folder name

mattclegg:


Otto42:
The theme folder name must be the URL sanitized version of the theme's name. So, perhaps it would be better to run the name through sanitize_title_with_dashes to get the folder name.

mattclegg:
I believe it will be a sanitized version.

'stylesheet' gets set here; https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-theme.php#L210

Otto42:
This seems incorrect to me. The "themename" variable you're adjusting here is the folder name. What problem does this fix, exactly?

Otto42:
The latest textdomain check notifies when the directory name is wrong for the theme name. If there are any further issues, let me know.

PR #76, opened 2015-07-30
Change tag from h2 to h1 in admin page

ernilambar:
Change tag from h2 to h1 in admin page


Otto42:
Fixed in another way instead.

PR #96, opened 2015-10-08
Avoid using HTML tags in translation strings

ramiy:
In the next version of WordPress (4.4), we fixed many translation strings that used html tags.

See: https://core.trac.wordpress.org/query?milestone=4.4&component=I18N&reporter=ramiy&group=milestone&col=id&col=summary&col=milestone&col=type&col=priority&col=component&col=severity&col=resolution&order=priority

The same thing should be done in this plugin. 

It's make it much easier to translates those strings. Removing the code, I can translate only the text.


Otto42:
Note that I updated the trunk version in the directory, so the strings should get pulled into translate w.org and will be available for translation in the Development (trunk) branch.

PR #108, opened 2016-01-01
updates

ramiy:


Otto42:
These appear to just be the same changes I already have. Please submit PRs for specific and targeted issues if there are any.

Issue #107, opened 2016-01-01
INFO for accessibility-ready tag


carolinan:
We briefly discussed an issue where some authors and reviewers don't know that there are additional guidelines for this tag. This is a question of education, but I believe theme check can be one way to do that.
A new info notice can be added together with a link to the guidelines.

Issue #106, opened 2016-01-01
Checking for disallowed theme names


carolinan:
When themes are submitted with a name that is not allowed, we normally ask them to rename it and submit it again, but this can be prevented to some degree.

Examples:
Names containing "theme", "WordPress"
Names containing default themes, "TwentyFifteen Child" and so on.

Otto42:
The uploader itself blocks the twenty-* names. Straight blocking this in theme-check could cause problems for any existing themes which pre-date this rule. Are there any such themes?

carolinan:
I wasn't able to search trac for names containing different versions of "theme".  There are several tickets that are closed as not approved, and others that are renamed, but the original ticket is still open so that they can keep their place in the que.  I did not find any live themes using it, but my search query was incomplete.I tried using Summary.. contains...THEME: combined with theme, themes, _theme etc.

carolinan:
OK actually this one is live: thememagic. https://themes.trac.wordpress.org/ticket/13982

grappler:
We can make the theme as a special case so that it can bypass the theme check but I don't expect will get any updates for as it was last updated three years ago.

@Otto42 DO you need any help adding the [check](https://github.com/Otto42/theme-check/pull/109/files#diff-936a7b276e8e6904cd670ef7ed388bb7R1)?

carolinan:
Surely there must be a better way to search for the names? The search I was able to do was incomplete, there might be more themes that are live.

grappler:
Here is a list of all of the themes with the name "theme". There are 47. Nearly all of them have not been updated since 2011. Some have not author and have normally less then 100 installs. There is only one theme with the word WordPress that is "IPhone WordPress Theme"
![image](https://cloud.githubusercontent.com/assets/1785641/13979077/216521d6-f0d6-11e5-9142-8f3c66fa8f41.png)

carolinan:
Would it be possible to allow it for updates but not for new themes?

grappler:
We could theoretically create a whitelist in the code, if need be, like with the blacklist. I wonder if those themes without an author should be even live. @Otto42 may have some ideas.

Otto42:
Patch seems reasonable. I'll take a look at it soon, tweak it a bit.

carolinan:
Here is an example of a theme that passed the current check:
https://themes.trac.wordpress.org/ticket/80585

These 3 themes have been updated within the last 2 years:
ithemer, themelia, themesbydp

Issue #105, opened 2016-01-01
Add more checks for auto generated themes


carolinan:
Currently the plugin is checking for themes generated by Artisteer.
This file artisteer.php should be renamed and include checks for other generators,

Known generators:
http://www.wpthemegenerator.com/ css prefix: wptg_

http://www.lubith.com/ .lubith file added, 
"* This theme was generated with Lubith - The Wordpress Theme Generator: http://www.lubith.com in file header.

http://templatetoaster.com/ css prefix: ttr_, php prefix: templatetoaster

Ref: https://wordpress.slack.com/archives/themereview/p1451336054004590

Issue #56, opened 2015-04-09
Theme URI and Author URI


benlumia007:
Theme URI and Author URI cannot be the same, which comes to.. 

users can bypass this if use a "/" at the end of a uri. 

theme uri: http://www.domainname.com
author uri: http://www.domainname.com/

emiluzelac:
Yep, as I reported in: https://wordpress.slack.com/archives/themereview/p1428536051003175

Otto42:
Seems a bit minor to me. But if you want to make a PR, I'll accept it.

Note that there is no current check to see if they're the same, that's one of the old hardcoded bits. Make a new check for it.

carolinan:
Are there any links besides Theme URI with WordPress.org that we should disallow?

Otto42:
You could add in the w.org domain as well, since it redirects.

carolinan:
Actually perhaps this should be combined with links.php in one file.

PR #109, opened 2016-01-02
Suggestions for fixes for, #107, #106, #105, #56

carolinan:
Suggestions for fixes for, #107, #106, #105, #56


Otto42:
These kind of large scale multi-part fixes all in one branch make it very difficult for me to merge them sanely. You need to use branches for each separate "thing" and then do a pull request for just that branch. I can then merge them piecemeal and individually instead of all at once.

Otto42:
Never mind, I can use cherry picking for this. In the future, try to use branches.

PR #112, opened 2016-01-12
Check not only PHP files but also CSS files also for hardcoded CDN URL

ernilambar:
Check not only PHP files but also CSS files also for hardcoded CDN URL. Modified `cdn.php`.


grappler:
@Otto42 Can we add this in so that https://github.com/Otto42/theme-check/issues/90 can be closed?

PR #113, opened 2016-02-07
Check for Private Functions

johnakos:
Check for private functions that are not intended for use by plugin and theme developers, but only by other core functions.

### DIFF
--- a/checks/private.php
+++ b/checks/private.php
@@ -1,0 +1,59 @@
+<?php
+
+class Access_Private implements themecheck {
+	protected $error = array();
+
+	function check( $php_files, $css_files, $other_files ) {
+		$grep = '';
+
+		$ret = true;
+
+		$checks = array(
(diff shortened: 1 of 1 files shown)


Otto42:
Why these functions in particular? Do we have a current issue where themes are using these that we need to prevent?

johnakos:
I think yes.
https://codex.wordpress.org/Category:Private_Functions

From wordpress.org codex

> 
This function's access is marked as private. That means it is not intended for use by plugin and theme developers, but only in other core functions. It is listed here for completeness

Otto42:
Yes, I get that. However, do you have any examples of any themes actually using these in a way that we need to detect for the review process?

Theme Check isn't meant to check everything possible, nor can it. It's a tool for our reviewers. It only aims to solve currently occurring problems.

Otto42:
Closing as there is no reason for this at this time.

PR #116, opened 2016-02-12
New check for @version line in each PHP template file

### LINKED ISSUE #114: Add dash to the regex looking for the theme slug
Is there any reason why the dash is missing in the regex? Theme slugs could contain a-z, 0-9 and dashes, so we could add that, or not?
See: https://github.com/Otto42/theme-check/blob/master/checkbase.php#L144

### LINKED ISSUE #115: Add new check for @version line in each PHP template file
I would like to suggest a new check for a doc block line starting with `@version` to give each template file an own version number (like [WooCommerce](http://speakinginbytes.com/2014/02/woocommerce-2-1-outdated-templates/))

With this `@version` line we can use the full power of this plugin:
https://wordpress.org/plugins/child-theme-check

I made this plugin last year. Here are my slides from my talk at WordCamp Cologne:
http://de.slideshare.net/zodiac1978/the-child-theme-dilemma-en

Zodiac1978:
Add dash to the regex to find the theme slug. Fixes #114
Add new check for @version line in each PHP template file. Fixes #115

### DIFF
--- a/checkbase.php
+++ b/checkbase.php
@@ -141,7 +141,7 @@ function tc_strxchr($haystack, $needle, $l_inclusive = 0, $r_inclusive = 0){
 }
 
 function tc_filename( $file ) {
-	$filename = ( preg_match( '/themes\/[a-z0-9]*\/(.*)/', $file, $out ) ) ? $out[1] : basename( $file );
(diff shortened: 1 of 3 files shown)


Otto42:
You need to separate these pull requests. I would merge the dash issue, but not the @version one at this time.

PR #117, opened 2016-02-25
Minor corrections and optimizations

gbaudhuin:
Make deprecated.php run much faster. On big themes, avoids crash due to php script maximum execution time error.


Otto42:
These all need to be separated into individual branches and made as individual pull requests, not as one big chunk of unrelated code.

I cannot accept this pull request at this time, in this form.

PR #135, opened 2016-03-23
Check for Text Domain in style.css - Fixes #130

### LINKED ISSUE #130: Include a text domain in style.css [Required]
Check style.css if `Text Domain:` is set correctly.

davidrgreen:
Fixes #130 I added a simple check for the presence of Text Domain: into the existing Style_Needed check. The other checks there checks that the line has the proper beginning, but doesn't validate the format of the rest of the line. This pull request mimics that, but could this be extended or is there a reason not to check the rest of the line?

If the rest of the line was checked, for checking the text domain it might be something more like
`[ \t\/*#]*Text Domain:\s*[A-Za-z0-9-]+\n`
where it confirms there is no underscore used. I haven't looked to see exactly how that's parsed yet so would do that before extending the regular expression.


grappler:
This is a good start. :+1: 

My ideas was also to check if the theme is using the the correct texdomain like in the check here: https://github.com/Otto42/theme-check/blob/master/checks/textdomain.php#L38

PR #137, opened 2016-04-05
Major strings cleanup and optimization

pedro-mendonca:
Remove non traslatable names from strings and transformed in variables.

Examples of removal: function_names() (and `<strong>`tags`</strong>`), links, html code like `<span class="tc-lead tc-warning">` , `<pre>code</code>`. reutilization of same strings and avoiding recundancy, crop out from the end of translation strings the variable that just renders the actual error list (ex. `$error`, `$grep`, etc),

This will avoid errors of information lost in translation for sensitive information like function names etc, reduce dramatically the number of strings with reapeated data (from 182 to 147!), makes strings easier to understand and translate.
General uniformization and simplification of translatable strings

Most of these changes won't make any changes in the final report shown, except those with small typos and those that were uniformized due to very small differences, for the better user experience.

A lot were similar, and some just reused the same sentences over and over, as the sensitive data was croped out as a variable outside of the string, huge ammount of similar strings obtained.

I did this as I've translated the plugin, so I checked it for coherence and usability.

### DIFF
--- a/checkbase.php
+++ b/checkbase.php
@@ -91,7 +91,7 @@ function tc_grep( $error, $file ) {
 			$error = ltrim( $error );
 		$pre = ( FALSE !== ( $pos = strpos( $this_line, $error ) ) ? substr( $this_line, 0, $pos ) : FALSE );
 		$pre = ltrim( htmlspecialchars( $pre ) );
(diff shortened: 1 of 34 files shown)


Otto42:
Some of these we'll implement, but removal of function names and emphasis HTML tags from the strings is not necessary nor desirable. Because there is so much change in this PR that will not be accepted, I'm going to close the whole thing and consider pieces of it individually instead.

pedro-mendonca:
Thank you, I'll make proper PRs with separate issues.
Regards

PR #147, opened 2016-04-22
Make text domain checks stricter

grappler:
As the TRT Requirement allows a maximum to two text domain, the check should fail if there are more than 2 text domains.

Now there is also a check that the correct text domain is used so to make use of langauge packs

### DIFF
--- a/checks/textdomain.php
+++ b/checks/textdomain.php
@@ -135,7 +135,7 @@ class TextDomainCheck implements themecheck {
 		$domainscount = count($domains);
 		
 		if ( $domainscount > 1 ) {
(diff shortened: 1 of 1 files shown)


Otto42:
You cannot remove the exception check here: https://github.com/Otto42/theme-check/pull/147/commits/d66911060d680f9c6fbb6177b464e683409ad5bc#diff-713127b7c50cea02c43dce0cf8c70f7bL35

Core themes have a different set of text domains than their name would normally indicate. This is a quirk specific to the core twenty-* themes.

grappler:
I moved the exception further down https://github.com/Otto42/theme-check/pull/147/files#diff-713127b7c50cea02c43dce0cf8c70f7bR127 I felt it was easier to manage all of the checks in one place.

PR #148, opened 2016-04-25
Title tag check changes

grappler:
This PR makes the use of `add_theme_support('title-tag')` required and not just recommended as it was released in WP 4.1 in December 2014.

This also changes from `<title>` tags and `wp_title()` being required to not allowed as `add_theme_support('title-tag')` has been added. `wp_title()` was nearly deprecated. http://wptavern.com/wordpress-4-4-to-deprecate-the-wp_title-function

### DIFF
--- a/checks/title.php
+++ b/checks/title.php
@@ -13,14 +13,15 @@ class Title_Checks implements themecheck {
 	function check( $php_files, $css_files, $other_files ) {
 		$ret = true;
 		$php = implode( ' ', $php_files );
-		
+
 		// Look for add_theme_support( 'title-tag' ) first
 		$titletag = true;
 		if ( ! preg_match( '#add_theme_support\s?\(\s?[\'|"]title-tag#', $php ) ) {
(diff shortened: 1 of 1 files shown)


grappler:
Includes a fix for the issue reported by @samikeijonen. 
https://wordpress.slack.com/archives/themereview/p1462473824002693

The problem was that the svg tag has a class in it. https://github.com/samikeijonen/some/blob/master/inc/template-tags.php#L127

Otto42:
Seems that your change in e0288e9 would cause a problem with the SVG usage there too, no?

grappler:
I just tested it with theme "some" and there was no issues but caught the `<title>` that I added in header.

We should really add some unit tests to the plugin :smile:

ernilambar:
@grappler May be rare case: When `<title><?php wp_title(); ?></title>` is there in `header.php` and in `functions.php` there is `add_theme_support( 'title-tag' );` there is no warning message regarding `<title>` tag.

Otto42:
That is basically the same as the previous behavior, and it is because of the && ! $titletag check in that if statement.

Issue #134, opened 2016-03-22
Don’t include any plugins [Required]


grappler:
Check for bundled zip files.

Possibly add a "Warning" check for `/* Plugin Name:`

PR #151, opened 2016-05-04
Adding check for included .zip files

kkoppenhaver:
Addresses #134 

Checks for and .zip files present in the repository and throws an error on finding one.

### DIFF
--- a/checks/included-plugins.php
+++ b/checks/included-plugins.php
@@ -6,14 +6,8 @@ class IncludedPlugins implements themecheck {
 		$ret = true;
 
 		$filenames = array();
-
-		foreach ( $php_files as $php_key => $phpfile ) {
-			array_push( $filenames, strtolower( basename( $php_key ) ) );
-		}
+		
 		foreach ( $other_files as $php_key => $phpfile ) {
-			array_push( $filenames, strtolower( basename( $php_key ) ) );
-		}
-		foreach ( $css_files as $php_key => $phpfile ) {
 			array_push( $filenames, strtolower( basename( $php_key ) ) );
 		}
 
--- a/checks/included-plugins.php
+++ b/checks/included-plugins.php
@@ -6,9 +6,9 @@ class IncludedPlugins implements themecheck {
 		$ret = true;
 
 		$filenames = array();
-		
-		foreach ( $other_files as $php_key => $phpfile ) {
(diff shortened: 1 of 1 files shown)


checks/included-plugins.php
```diff
@@ -0,0 +1,40 @@
...
+		$blacklist = array(
+			'\.zip'	=> __( 'Zipped Plugin', 'theme-check' ),
+		);
+
+		checkcount();
+
+		foreach( $blacklist as $file => $reason ) {
+			if ( $filename = preg_grep( '/' . $file . '/', $filenames ) ) {
+				$error = implode( array_unique( $filename ), ' ' );
+				$this->error[] = sprintf( '<span class="tc-lead tc-required">'.__( 'REQUIRED','theme-check').'</span>: '.__( '<strong>Zip file found.</strong> Plugins are not allowed in themes. The zip file found was ', 'theme-check') . '<em>' . __( $error ) . '</em>.' ) ;
```
grappler: We can't translate variables `__( $error )`. As you are already using `sprintf` you can do `__( 'The zip file found was %s', 'theme-check' )` and then define `$error` as the counterpart to `%s`

kkoppenhaver:
Good call, updated.

Otto42:
You don't need to check the $php_files and $css_files variables. Those are literally just *.php and *.css. They can't be zip files.

kkoppenhaver:
Thanks, updated.

Otto42:
Also, you have this as a REQUIRED check. Has this been discussed amongst the theme-review team? Is this requirement documented in the theme review documentation somewhere?

If not, then it needs to be a RECOMMENDED check instead. Theme Check only implements the existing requirements as created by the review team, it doesn't create new ones. That said, it would be fine as recommended, because including ZIPs in themes makes little sense.

grappler:
Yes, we discussed this before. The requirement this is based on is https://make.wordpress.org/themes/handbook/review/required/#plugins

It does not make sense to have additional zip files in the theme.

kkoppenhaver:
Thanks everyone for the tips and pointers. This was my first PR!

grappler:
@kkoppenhaver I just wanted to let you know this PR has been a success. I has caught a number of themes who included plugins in the theme.

kkoppenhaver:
@grappler That's awesome to hear!

PR #156, opened 2016-05-10
Add dash to regex and fix typo (Fixes #114)

### LINKED ISSUE #114: Add dash to the regex looking for the theme slug

Zodiac1978:
I'm not a Git guru, so I just reverted the unnecessary commit. Fine? Or what should I do to split this up?


Otto42:
Generally speaking, you make a new "branch" for each separate "thing" that you're working on. This is because when you submit a PR, then you're submitting all the changes in that whole branch. Updating the branch later updates the PR as well. Thus, each branch should consist of one whole change for one specific purpose.

Reverting the other change is fine at this time. For the future, just so you know, when you want to work on an unrelated thing, switch back to the master, then branch from there for the new thing to be worked on. Keeps changes separate. :)

Zodiac1978:
Yes, that was may fault. I should have created a new branch for that. Thanks for the merge.

Unfortunately the title of the reverted commit still triggered the close of #115 - could you re-open the ticket for me @Otto42 ? Thanks in advance!

PR #159, opened 2016-05-24
add ' and " to lookback in include regex.

Pross:
Fixes https://github.com/Otto42/theme-check/issues/158


Pross:
https://regex101.com/r/rF8zO5/2

Issue #157, opened 2016-05-18
Unnecessary add_theme_support( 'menus' )


grappler:
I have seen a number of theme use `add_theme_support( 'menus' )` which is unnecessary as `register_nav_menus` does everything.

grappler:
Fixed by #161

PR #161, opened 2016-06-05
Add check for redundant menu support declaration code

ernilambar:
See #157


grappler:
Please could you mention that [register_nav_menus()](https://developer.wordpress.org/reference/functions/register_nav_menus/) should be used instead.

> Reference to <strong>add_theme_support( "menus" )</strong> was found in the theme. This should be  removed and register_nav_menus() used instead.

It would be better to split the sentences and the replace the function name with `%s`. This will make it easier to translate.

grappler:
Thanks, could you also move the strong tags out of the translation as they do not need to be translated?

ernilambar:
@grappler `strong` tag is moved out from the string.

Issue #152, opened 2016-05-06
Info message for old way of Custom Header Implementation


ernilambar:
I have seen some constants being used in some theme. Eg, `HEADER_IMAGE_WIDTH`, `HEADER_IMAGE_HEIGHT`. It would be great if we add some `INFO` message to encourage implementing `add_theme_support( 'custom-header' )`

ernilambar:
Following are the `define` variables which are deprecated.

* NO_HEADER_TEXT
* HEADER_IMAGE_WIDTH
* HEADER_IMAGE_HEIGHT
* HEADER_TEXTCOLOR
* HEADER_IMAGE
* BACKGROUND_COLOR
* BACKGROUND_IMAGE

@grappler What if we make it required and fail process if such define variables are used in the theme. They are deperecated since long time.

grappler:
Yes, I would be fine with that due to this comment in core:
> // Constants are lame. Don't reference them. This is just for backwards compatibility.

ernilambar:
Fixed in https://github.com/WordPress/theme-check/pull/162

PR #162, opened 2016-06-08
Add check for deprecated define variables

ernilambar:
See #152 

**Current limitation:**

 Theme's functions.php
```
define('HEADER_IMAGE_WIDTH',1200);
define('HEADER_IMAGE_HEIGHT',600);
```

Output:
```
REQUIRED: HEADER_IMAGE_WIDTH found in the file functions.php. Deprecated since version 3.4. Use add_theme_support( 'custom-header' ) instead.
Line 39: define('HEADER_IMAGE_WIDTH',1200);
REQUIRED: HEADER_IMAGE_HEIGHT found in the file functions.php. Deprecated since version 3.4. Use add_theme_support( 'custom-header' ) instead.
Line 40: define('HEADER_IMAGE_HEIGHT',600);
REQUIRED: HEADER_IMAGE found in the file functions.php. Deprecated since version 3.4. Use add_theme_support( 'custom-header' ) instead.
Line 39: define('HEADER_IMAGE_WIDTH',1200);
Line 40: define('HEADER_IMAGE_HEIGHT',600);
```
Although there is no `HEADER_IMAGE` variable, regex matches with `HEADER_IMAGE_WIDTH`. Any suggestion how to tighten the regex check?


checks/deprecated-define.php
```diff
@@ -0,0 +1,59 @@
...
+class Deprecated_Define implements themecheck {
+	protected $error = array();
+
+	function check( $php_files, $css_files, $other_files ) {
+		$grep = '';
+
+		$ret = true;
+
+		$checks = array(
+			array( "NO_HEADER_TEXT" => "add_theme_support( 'custom-header' )", '3.4' ),
```
grappler: The WP coding standards is to use single quotes in array instead of double.

grappler:
It may be easier to extend an existing class [constants.php](https://github.com/Otto42/theme-check/blob/master/checks/constants.php)

checks/deprecated-define.php
```diff
@@ -0,0 +1,59 @@
...
+			array( "HEADER_TEXTCOLOR" => "add_theme_support( 'custom-header' )", '3.4' ),
+			array( "BACKGROUND_COLOR" => "add_theme_support( 'custom-background' )", '3.4' ),
+			array( "BACKGROUND_IMAGE" => "add_theme_support( 'custom-background' )", '3.4' ),
+		);
+		foreach ( $php_files as $php_key => $phpfile ) {
+			foreach ( $checks as $alt => $check ) {
+				checkcount();
+				$key = key( $check );
+				$alt = $check[ $key ];
+				if ( preg_match( '/' . $key . '/', $phpfile, $matches ) ) {
```
grappler: You could try `preg_match( '/(?:'|")' . $key . '(?:'|")/', $phpfile, $matches )`

grappler:
It may be easier to extend an existing class [constants.php](https://github.com/Otto42/theme-check/blob/master/checks/constants.php)

PR #165, opened 2016-06-15
The use of #wpadminbar should not fail the check

### LINKED ISSUE #164: You are not allowed to hide the admin bar when wpadminbar is used.
So `REQUIRED: You are not allowed to hide the admin bar.` will show if author is using `#wpadminbar` ID in the stylesheet. 

I don't think that this is needed. There are few scenarios where authors will need `#wpadminbar`, like wall-to-wall images with an absolute position, featured images with the same, etc.

https://github.com/Otto42/theme-check/issues/127

grappler:
fixes #164

### DIFF
--- a/checks/adminbar.php
+++ b/checks/adminbar.php
@@ -29,9 +29,8 @@ class NoHiddenAdminBar implements themecheck {
 				$filename = tc_filename( $file_path );
 
 				if ( preg_match( $css_regex, $file_content, $matches ) ) {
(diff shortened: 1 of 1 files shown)


amindiary:
seems not fixed yet! I still get this error! because I used 
```
#wpadminbar {
    z-index: 500!important;
    top:0;
}
```

shivampaw:
True ^

I set mine to fixed all the time but it still shows as a fail in ThemeCheck :(

Issue #167, opened 2016-06-21
Check for translation-ready(Required tag), if it's not there give warning


juhi123:
Theme check should check for translation-ready (required) tag and if it's not there it should give warning to add this tag
In this link https://make.wordpress.org/themes/handbook/review/required/theme-tags/, translation-ready tag is being shown as required.

Otto42:
Continue on #168

PR #168, opened 2016-06-21
Issue #167

juhi123:
Added pull request for the same.


grappler:
Thank you for the PR. The translation-ready tag is not required but the theme is required to translation-ready. If all themes are translation-ready then the tag could be removed.

juhi123:
@grappler  Thanks for clarifying :) then I think the documentation should be updated. Because I have always thought this tag is required to be added :laughing:

carolinan:
Closing for now since the tag is not required,

PR #170, opened 2016-07-14
Revert "The use of #wpadminbar should not fail the check"

watchlondon:
Reverts Otto42/theme-check#165


Otto42:
Negative, this change was agreed to by the themes team. Changes along these lines should be discussed first.

Issue #172, opened 2016-07-16
warrning with echoing short tags


xandros15:
I got error `WARNING: Found PHP short tags in file` `Line 20: <a class='<?= $tab['class'] ?>'`.
I know using the normal short tags `<?` is bad practice, 'cuz colision with xml, 
but after php 5.4 `<?=` is a safer to use, and makes cleaner template script than `<?php echo`.

grappler:
We will be updating this as we move over to using PHP code sniffer. There is pull request open to check for alterntive tags. https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/585

I am currently working on the [new theme check](https://github.com/WPTRT/WordPress-Coding-Standards) instead of fixing issues here.

You can create all pull request if you want. The check is found in `phpshort.php`

Otto42:
Why would we allow short echo's? Those should not be used. They don't work in PHP < 5.4.

The problem with short echo is that it makes it too easy to accidentally output unsanitized data, potentially leading to XSS vulnerabilities. At least requiring the use of "echo" makes it obvious when something is being output to the page. Just doing <?= $variable ?> does not make it as obvious that the variable is being output or needs to be sanitized properly, at least not as much as <?php echo $variable ?> does.

My vote would be to not allow use of short echo in themes at all. It makes no sense to introduce potential security mistakes just to save a few keystrokes.

xandros15:
> They don't work in PHP < 5.4.

Still don't know why wordpress supports php lower than 5.4 (even lower than 5.3)
But the worsted thing is usage PHP 4 > PHP 7
(source https://w3techs.com/technologies/details/pl-php/all/all)
> too easy to accidentally output unsanitized data, potentially leading to XSS

wat? `<?php echo esc_html($var) ?>` vs `<?= esc_html($var) ?>`... so we should use `array()` too?
Views should only print data, no manipulate on it. The most of variables in views will be echoing.

Not my project, so I will don't insist.

Ehsan-Dalvand:
See this answer: 
http://stackoverflow.com/questions/14188397/why-are-echo-short-tags-permanently-enabled-as-of-php-5-4#answer-14188438

> Short open tags are not always enabled since PHP 5.4. The documentation talks about the short echo tags. Which is a different thing....

carolinan:
We decided to keep the existing check with a warning for the echo short tag.
[...]

PR #173, opened 2016-07-16
#172 - add echo short tag support

xandros15:


grappler:
There is no need to add a constant to enable the echo short tag. You can directly allow it.

xandros15:
I prefer to control some things by constants (or shows that option can be change), but if you insist...

Otto42:
Why would we allow short echo's? Those should not be used. They don't work in PHP < 5.4.

pattonwebz:
Since core php recommendation is now 5.6 and that in 5.4 short echo tags were decoupled from the [short_open_tag](https://www.php.net/manual/en/ini.core.php#ini.short-open-tag) setting the only reason I see that we should continue to disallow these is for sake of consistency - core does not use short echo tags anywhere.

My personal preference would be to continue using the full syntax but there is no longer a technical reason to keep these in the disallow regex.

dingo-d:
The WPCS clearly states not to allow them: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#no-shorthand-php-tags.

While I agree that themes shouldn't follow WPCS, should we try to be as consistent as possible with them?

joyously:
For me, this is one of those things that falls into "we don't have just any theme, we have the best themes" sentiment. Themes should be readable, and short echo tags are not (for the average coder).
I would keep them in the disallow regex for consistency with core and legacy themes.

carolinan:
For consistency, we will not add support for this, so, closing.

PR #185, opened 2016-12-18
This check fails - it should be marked as required

grappler:
As `$ret` is set to `true` this will fail the check. The text should show that too by marking as required and not just info.


carolinan:
We may need to revert this until we have solved https://github.com/WordPress/theme-check/issues/221

PR #188, opened 2017-04-07
Design Fixes

maheshwaghmare:


GaryJones:
Could you give before and after screenshots, so we can see the difference?

maheshwaghmare:
Hi @GaryJones,

This is the minor CSS change. Check the difference below:

<table>
<tr><th>Before</th><th>After</th></tr>
<tr><td></td><td></td></tr></table>

PR #198, opened 2018-06-23
Fix typo

Rub3nCT:
Add missing whitespace. [REQUIRED: This theme text domain does not match the theme's slug. The text domain used: XXXThis theme's correct slug and text-domain is XXX.]


Otto42:
This seems to be incorrect, as it adds a period inside the <strong> element.

PR #205, opened 2019-05-17
allow for multiple style.css files

joyously:
Move the checks into the loop.
Output the folder name when Tags are missing.

The master is missing the 2 new Tags wide-blocks and block-styles, but the version I got from the .org repo had them, so they are in this PR.

### DIFF
--- a/checks/style_tags.php
+++ b/checks/style_tags.php
@@ -9,53 +9,57 @@ class Style_Tags implements themecheck {
 		$filenames = array();
 
 		foreach( $css_files as $cssfile => $content ) {
-			if ( basename( $cssfile ) === 'style.css' ) $data = get_theme_data_from_contents( $content );
-		}
+			if ( basename( $cssfile ) === 'style.css' ) { 
+				$data = get_theme_data_from_contents( $content );
 
-		if ( !$data[ 'Tags' ] ) {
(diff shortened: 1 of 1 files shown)


carolinan:
@joyously Can you explain the issue? We are are not sure what problem this solves.

Please resolve the conflicts above.

joyously:
Sorry I didn't create an issue. 
At the time that I wrote the 4 PRs, I took the latest version from the plugin, which already contained the new Tags for blocks, but those Tags were not in the code on GitHub.

It is possible for a theme to contain resources that have their own style.css (like in a subfolder). Theme Check will still check this file as if it is the main style.css and then stop checking other ones.
This PR moves the checks into the loop, so that all style.css are checked, and the error for Tags will output the folder name of the style.css file, so it is more obvious which one the error refers to.

dingo-d:
Should we prevent theme-check even checking the other `style.css` files (the one not in the root) in the first place. That is the one that the core (and .org) will use to pick up the information about the theme. So why wasting resources by checking other ones?

joyously:
The way it is now, it checks the first one it finds and gets out. That isn't always the one in the root.

dingo-d:
Yeah, this will need to get fixed :/

carolinan:
-Works well in my tests. Perhaps the folder name could be bold _or something_ so that it is more visible.

PR #222, opened 2020-03-16
Plugin Territory - Actions & Filters

acosmin:
This PR adds a few actions and filters I recently encountered in a few of my reviews. 

Filters: 
* `mime_types`, `upload_mimes`
* `user_contactmethods`

Actions:
* `wp_dashboard_setup`, ([example](https://themes.trac.wordpress.org/browser/business-lander/1.1.11/inc/dashboard-widget.php#L12))

Can be tested by adding this in a theme:
```
add_action( 'wp_dashboard_setup', function() {} );
add_filter(  "upload_mimes" ,  function( $mimes ) { return $mimes; } );
add_filter ('user_contactmethods',function( $methods ) { return $methods; } );
```

### DIFF
--- a/checks/plugin-territory.php
+++ b/checks/plugin-territory.php
@@ -34,6 +34,28 @@ class Plugin_Territory implements themecheck {
 			$ret = false;
 		}
 
+		// Hooks (actions & filters) that are required to be removed from the theme.
+		$forbidden_hooks = array(
+			'filter' => array(
+				'mime_types',
+				'upload_mimes',
+				'user_contactmethods'
+			),
+			'action' => array(
+				'wp_dashboard_setup'
+			)
+		);
+
+		foreach ( $forbidden_hooks as $type => $hooks ) {
+			foreach ( $hooks as $hook ) {
+				checkcount();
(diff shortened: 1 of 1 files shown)


joyously:
I'm wondering about the regex for the filters and actions. An author could easily get around that by making their own function, so then the hook is a variable.

Also, what do you think about adding `update_option` to the list of functions? Since themes must use the Customizer, which does the update_option for them, they shouldn't be calling that function.

acosmin:
True, I guess you could bypass it that way, but you could probably bypass most of these checks. We're still required to check each file, line by line. You're probably going to notice it.

I wouldn't add `update_option` because if you use type `option` instead of `theme_mod`, then you probably need a way to update that option outside the Customizer (even if all the modifications are inside one single option/array). There are different use cases for it.

joyously:
This is unrelated to the PR, but since the comment is here, I want to answer it:
> I wouldn't add `update_option` because if you use type `option` instead of `theme_mod`, then you probably need a way to update that option outside the Customizer (even if all the modifications are inside one single option/array). There are different use cases for it.

The whole point is that themes are required to use a single array and the Customizer, so it shouldn't be in the theme. I know some plugins that get bundled into themes may have a call to `update_option`, but I don't think they should, due to the requirements I just mentioned.

PR #225, opened 2020-03-26
Add check for disallowed theme names

carolinan:
Solves #106 

-Now allows exceptions for existing themes with the words theme and WordPress in the names.

### DIFF
--- a/checks/themename.php
+++ b/checks/themename.php
@@ -1,0 +1,56 @@
+<?php
+/**
+ * Check for prohibited words and theme names.
+ */
+
+
+class ThemeNameCheck implements themecheck {
+	protected $error = array();
+
+	function check( $php_files, $css_files, $other_files ) {
+
+		checkcount();
+		$ret = true;
+		global $themename;
+
+		/* Get the author name from style.css. */
+		foreach ( $css_files as $cssfile => $content ) {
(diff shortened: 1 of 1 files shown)


checks/themename.php
```diff
@@ -0,0 +1,49 @@
...
+		/* If the author is not the WordPress team, check for prohibited names. */
+		if ( 'the WordPress team' !== $author ) {
+
+			$blacklist = array( 'twentyten', 'twentyeleven', 'twentytwelve', 'twentythirteen', 'twentyfourteen', 'twentyfifteen',
+			'twentysixteen', 'twentyseventeen', 'twentyeighteen', 'twentynineteen', 'twentytwenty', 'twentytwentyone',
+			'twentytwentytwo', 'twentytwentythree', 'twentytwentyfour', 'twentytwentyfive', 'twentytwentysix',
+			'twentytwentyseven', 'twentytwentyeight', 'twentytwentynine', 'twentythirty'
+			);
+
+			foreach ( $blacklist as $key ) {
```
dingo-d:
```suggestion
			foreach ( $blacklist as $theme_name ) {
```

Technically it's array value, but it's always better to have a clear naming. So `$theme_name` fits better imo.

carolinan:
The check needs to be updated to find theme names like mb_theme.

Issue #212, opened 2019-07-23
Add warnings for URLs to image sites with non GPL images.


carolinan:
There are websites that offer non gpl compatible images, that are commonly used in themes,
a warning should be added for these:
unsplash.com
pixabay.com

The reason why this needs to be a warning and not an error is that older themes may still have 
images from these sites.

carolinan:
https://make.wordpress.org/themes/2017/06/30/theme-review-meeting-summary-june-27-2017/

PR #226, opened 2020-03-26
Check for URLs to image sites with non GPL images

carolinan:
Checks if the theme includes resources from websites that does not use a GPL compatible license.
#212

**This must not be merged until we have a decision whether it needs to be a warning or a requirement.**

### DIFF
--- a/checks/nongplsites.php
+++ b/checks/nongplsites.php
@@ -1,0 +1,47 @@
+<?php
+/**
+ * Checks if the theme includes resoruces from websites that does not use a GPL compatible license.
+ */
+class NonGPLCheck implements themecheck {
+	protected $error = array();
+
+	function check( $php_files, $css_files, $other_files ) {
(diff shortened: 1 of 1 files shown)


checks/nongplsites.php
```diff
@@ -0,0 +1,46 @@
...
+
+	function check( $php_files, $css_files, $other_files ) {
+
+		$ret  = true;
+		$code = implode( ' ', $other_files ); // The references are usually in the readme.
+
+		checkcount();
+
+		$link_list = array(
+			'unsplash'            => 'https://unsplash.com/license',
```
dingo-d: Maybe we should remove protocol, so if `http` or `https` is uses this would catch it, no?

carolinan: The url is not used in the check, it is only so that the user can go there to read the license.

Issue #123, opened 2016-03-22
No removing or modifying non-presentational hooks [Required]


grappler:
We can best solve this with a check for `remove_action(‘wp_head’)`

grappler:
I would use the following regex `remove_action\(.*("|')wp_head("|').*\);` It should not have any false positives. 

You can see it in action here: https://regex101.com/r/gV7jB2/2

PR #231, opened 2020-04-02
Check for remove_action wp_head

carolinan:
Solves #123

### DIFF
--- a/checks/plugin-territory.php
+++ b/checks/plugin-territory.php
@@ -51,9 +51,16 @@ class Plugin_Territory implements themecheck {
 				checkcount();
 				if ( preg_match( '/[\s?]add_' . $type . '\s*\(\s*([\'"])' . $hook . '([\'"])\s*,/', $php ) ) {
 					$this->error[] = '<span class="tc-lead tc-required">' . __( 'REQUIRED', 'theme-check').'</span>: ' . sprintf( __( 'The theme uses the %1$s %2$s, which is plugin-territory functionality.', 'theme-check' ), '<strong>' . esc_html( $hook ) . '</strong>', esc_html( $type ) );
-					$ret = false;
+					$ret           = false;
 				}
 			}
+		}
+
+		// Removal of some actions and filters are not allowed, this can be extended later when we have a larger list.
+		if ( preg_match( "/remove_action\(.*(\"|')wp_head(\"|').*\);/", $php ) ) {
+			checkcount();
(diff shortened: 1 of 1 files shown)


carolinan:
This needs an exception for moving the output of the custom CSS.
This is needed when you have inline CSS.
Example:

`/**
 * Custom CSS should be at the end of everything else in order to override existing styles.
 * We need to unhook it from wp_head and hook it in wp_footer.
 */
add_action(
	'wp_head',
	function() {
		remove_action( 'wp_head', 'wp_custom_css_cb', 101 );
		add_action( 'wp_footer', 'wp_custom_css_cb', PHP_INT_MAX );
	},
	10
);`

joyously:
That makes no sense. You don't put styles in the footer. You just use a large priority, so it comes out last, and that's what the default does.
Themes should not mess with `wp-head`.

aristath:
@joyously it's not unheard of, and it's not bad practice. It's unusual and not the way we - in WordPress - are used to doing things, but there's lots of benefits to loading non-critical styles in the footer. 
If a theme follows best web performance practices, they will load some styles in the head and others in the footer, or even inline. Styles not in head get loaded asynchronously without blocking the content rendering and are very much appreciated on slow connections because the user gets the content faster.
It's reasonable to allow moving custom-css to the footer.

joyously:
I still don't think themes should be moving actions from core. A plugin could be doing something else with it, each style sheet can have dependencies, and there is dynamic CSS from theme and plugins.

joyously:
Yes, I am aware. But plugins can do whatever, so it could be different than standard. The theme has no business moving it. It is not the theme's CSS to be moving.

dingo-d:
I don't get the issue here, this will just check if the theme removes this action, which we do not allow. So why is having this check a bad thing? 🤷‍♂️

aristath:
There are many cases that would require moving the custom-CSS to the footer... For example if a theme only prints styles for active blocks on a page (instead of enqueueing a single large stylesheet with all block-styles in `head`), then they need to detect which blocks are included in the content. That can't happen before the content renders, which means that block styles may be printed in the footer. Custom-CSS needs to override all other styles which means it will have to be printed _after_ the block styles, so it will need to be moved from head to the footer.
There are many more examples of valid uses cases, this is just the 1st one that came to mind.
Removing the action is bad only if it doesn't get added back again...

carolinan:
I mentioned this is in the theme review slack but I am not sure if it would be better to use a black list here and specify what we don't allow. We don't have much data right now on what themes are using (or why) .

Issue #163, opened 2016-06-09
You are not allowed to hide the admin bar.


wordica:
REQUIRED: You are not allowed to hide the admin bar.

I do not use in my theme any function to hide admin bar. I just check like this:

$check_a_ = get_user_meta(get_current_user_id(),'show_admin_bar_front');

if admin is on by user or not. Does this create this error?

grappler:
I don't think the theme check is checking for `show_admin_bar_front`. Do you have the link to the themes code?

Where are you submitting the theme?

wordica:
I can't put link but for sure I don't do nothing to hide admin bar - I only check if it's visible or not.

wordica:
Theme-check has bug - it shows error when someone have #wpadminbar{} in CSS.

grappler:
It is not a bug but a deliberate check to try to spot themes that are hiding the admin bar with CSS: The check has been changed from required to warning. https://github.com/Otto42/theme-check/pull/165

flynn723:
This is still required, when I use theme checker.

AshekurRahman:
How to solve this problem. same to me.

wordica:
Right now I use jQuery to check if user use admin bar or not and it solve problem with theme check

danyj:
it is also triggered with this 

`#wpadminbar .quicklinks .{... any property`

carolinan:
It seems only one of the checks were changed to a warning, so this needs to be updated.

PR #232, opened 2020-04-02
Warn for hiding the adminbar

carolinan:
Warn for hiding the adminbar. Also adds file names and line numbers.
Setting this to a warning instead of required lets theme authors check if the admin bar is visible or not
(Useful for example to change position of elements below the bar).
Returns true, so that upload is not blocked.

Solves #163

### DIFF
--- a/checks/adminbar.php
+++ b/checks/adminbar.php
@@ -6,11 +6,12 @@ class NoHiddenAdminBar implements themecheck {
 	protected $error = array();
 
 	function check( $php_files, $css_files, $other_files ) {
-		$ret = true;
-		checkcount();
+		$ret      = true;
+
 		$php_regex = "/(add_filter(\s*)\((\s*)(\"|')show_admin_bar(\"|')(\s*)(.*))|(([^\S])show_admin_bar(\s*)\((.*))/";
 		$css_regex = "/(#wpadminbar)/";
 
+		checkcount();
 		// Check php files for filter show_admin_bar, show_admin_bar_front, and show_admin_bar().
 		foreach ( $php_files as $file_path => $file_content ) {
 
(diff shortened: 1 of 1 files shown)


checks/adminbar.php
```diff
@@ -1,41 +1,48 @@
...
+			$error    = '/#wpadminbar/';
+			$grep     = tc_preg( $error, $file_path );
 
-				if ( preg_match( $css_regex, $file_content, $matches ) ) {
-					$this->error[] = sprintf( '<span class="tc-lead tc-warning">' . __( 'WARNING', 'theme-check').'</span>: ' . __( 'The theme is using `#wpadminbar`. Hiding the admin bar is not allowed.', 'theme-check' ), 
-						'<strong>' . $filename . '</strong>');		
-				}
+			if ( preg_match( $css_regex, $file_content, $matches ) ) {
+				$this->error[] = sprintf( '<span class="tc-lead tc-warning">' . __( 'WARNING', 'theme-check' ) . '</span>: ' . __( 'The theme is using `#wpadminbar` in %1$s. Hiding the admin bar is not allowed. This warning must be manually checked.', 'theme-check' ),
+				'<strong>' . $filename . '</strong>' ) . $grep;
```
carolinan, line 38: $grep will output the entire line, so, it should't be output if the CSS file is minified.

dingo-d: You can check for minified files like I did in the theme sniffer: https://github.com/WPTRT/theme-sniffer/blob/dcfd2f181fae0ea17c99e461ebf2bc7be571e64f/src/callback/class-run-sniffer-callback.php#L424

One additional check that comes to mind is to 'parse' a file through `file_get_contents()` function, and get the line length. Since minified files are usually 1 line, but super long, that could be an extra check (if line length longer than 200-300 characters skip).

PR #246, opened 2020-04-18
WIP Add checks for common escaping issues

carolinan:


checks/escaping.php
```diff
@@ -0,0 +1,45 @@
...
+	protected $error = array();
+
+	function check( $php_files, $css_files, $other_files ) {
+
+		$ret = true;
+
+		$checks = array(
+			'/echo get_theme_mod/'              => __( 'Theme options must be escaped.', 'theme-check' ),
+			'/="<\?php esc_html_e/'             => __( 'Use esc_attr_e() inside HTML attributes.', 'theme-check' ),
+			'/="<\?php echo esc_html__/'        => __( 'Use esc_attr__() inside HTML attributes.', 'theme-check' ),
```
dingo-d, line 16: How do you distinguish when this needs to be triggered? It's not that easy to check attributes. For instance, `href` is an attribute, but if it's populated via variable it should be escaped using `esc_url`. Also, these are for translatable escaping. 

I'm just worried this could trigger false positives 🤷‍♂️

carolinan: I share the concern but I also think the simpler the better...

carolinan:
> What about the use of `print` instead of `echo`?

I would like to check for that but not sure what the easiest way would be, since the content might be in the middle of a text, for example.

carolinan:
Changed the check for echo get_theme_mod to a warning to allow the short hand if statements that checks if an option is enabled and then echoes a specific value that is not user input.

Issue #217, opened 2020-01-08
Ignore build tools directories when performing the checks.


stefanobartoletti:
I think it would be useful to ignore certain directories related to built tools when scanning theme files.

Some default ignores could be `node_modules`, `vendor`, `.git` and so on, but maybe also give the user the ability to manually select them.

joyously:
Everything hosted in the repository must be checked.
Not everything from the source needs to be included in the final zip file, though. Evertyhing submitted must be reviewed.

stefano-b:
These directories won't go in the repository, they are only used by local build tools like `npm`, `gulp`, `composer` and so on. They are used only during the development phase.

joyously:
This plugin is the same code that is used for themes being uploaded to the repository. The folders in the theme folder being checked should match what will be uploaded.
In other words, you can't be testing correctly if your theme folder contains other files that won't be in the final zip.

pattonwebz:
As per my comment in #204 I am hesitant to exclude these directories from the scan because if they are in the theme they should be checked at upload time. None of the themes in the directory currently have large enough dev directories to fatal the scanner or else the upload check would be throwing fatal errors and I have not heard any reports about that.

Perhaps one idea could be to add a filter flag that would allow a developer to exclude those directories during local development? What do you think of that idea?

pattonwebz:
I opened PR #268 to add filters that can allow developers to skip over checking other files in a few directories while they are building the theme. These files would still be checked via the theme uploader on .org since the filters will not be in use there.

stefano-b:
My idea was exactly to exclude folders used during development, that won't go in the package in any circumstance. Adding these filters seems a nice solution.

PR #268, opened 2020-08-02
Add filters to skip checking other files in common dev directories

### LINKED ISSUE #204: Fatal error: Allowed memory exhausted
many themes when check end up with error "Fatal error: Allowed memory size of 268435456 bytes exhausted (tried to allocate 141901616 bytes) in /wordpress/wp-content/plugins/theme-check/main.php on line 30"

pattonwebz:
This adds a pair of filters that can be used to exclude checking of other files (non `.php` and `.css` files) during local development. It helps prevent fatal from memory exhaustion caused when the `node_modules` or `vendors` directory has lots of developer dependancies.

`tc_skip_development_directories`
`tc_common_dev_directories`

Fixes: #204 #217

### DIFF
--- a/main.php
+++ b/main.php
@@ -31,6 +31,13 @@ function check_main( $theme ) {
 			} elseif ( substr( $filename, -4 ) === '.css' && ! is_dir( $filename ) ) {
 				$css[ $filename ] = file_get_contents( $filename );
 			} else {
+				// In local development it might be useful to skip other files
+				// (non .php or .css files) in dev directories.
+				if ( apply_filters( 'tc_skip_development_directories', false ) ) {
+					if ( tc_is_other_file_in_dev_directory( $filename ) ) {
+						continue;
+					}
+				}
 				$other[ $filename ] = ( ! is_dir( $filename ) ) ? file_get_contents( $filename ) : '';
 			}
 		}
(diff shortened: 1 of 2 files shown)


main.php
```diff
@@ -224,3 +231,29 @@ function tc_form() {
...
+ */
+function tc_is_other_file_in_dev_directory( $filename ) {
+	$skip     = false;
+	// Filterable List of dirs that you may want to skip other files in during
+	// development.
+	$dev_dirs = apply_filters(
+		'tc_common_dev_directories',
+		array(
+			'node_modules',
+			'vendor',
```
dingo-d, line 249: I wouldn't add `vendor` here by default, as it can be needed in production builds sometimes. `node_modules` can always be excluded.

pattonwebz: In many themes I've worked on the `node_modules` directory is also sent in production builds. `npm --save` vs `npm --save-dev`.

I should update the name of the filters to make it clearer what they do because they don't technically skip the directories even when the filters are used. The only cause it to skip files that are not `.php` or `.css` files. So even if /vendors/ dir is excluded via the filter it will still scan the .php files from there :)

In my testing the problem is not actually caused by scanning of importat files for the checks here - it's parsing of the binary file types in these directories which exhausts memory or hits timeout limits.

dingo-d: Oh wow, node_modules should really never be in the production :S usually all js is just bundled and used in a project.
I can see the merit of completly skipping both folders entirely for theme check purposes (not to exhaust the memory)

pattonwebz: @dingo-d any chance you could be my 2nd set of eyes and test this with me? I have tested myself already with over a dozen different themes I had in my local but I always like to have an extra person look it over incase I missed anything.

dingo-d: I've tested it with the company's boilerplate and it seems to be working. When I put `add_filter('tc_skip_development_directories', '__return_true')` in my themes `functions.php` the check works and I can see

```
WARNING: The theme is using `#wpadminbar` in node_modules/@wordpress/edit-post/build-style/style.css. Hiding the admin bar is not allowed. This warning must be manually checked.Line 56: body.js.is-fullscreen-mode #wpadminbar {
```

but without it, I get memory exhausted error on `theme-check/checks/nongplsites.php on line 11`.

When I added 

```php
add_filter( 'tc_common_dev_directories', function( $ignored_folders ) {
  $ignored_folders[] = '.storybook';

  return $ignored_folders;
} );
```

I could see that folder ignored from the 

```
REQUIRED: .phpstorm.meta.php .eslintrc .stylelintrc .gitignore .eslintignore .travis.yml Hidden Files or Folders found. This file must not be in a theme.
```

error. So all seems to be ok :+1:

dingo-d (approved):
LGTM

PR #270, opened 2020-08-08
Add Check for Affiliate Links

aaronjorbin:
Some basic checks for known affiliate link code. Inspired by some of the code in https://github.com/brainstormforce/astra/commit/b3f7369eaaae128b451c2021bfcd8b8c3aaec45d

### DIFF
--- a/checks/badthings.php
+++ b/checks/badthings.php
@@ -17,6 +17,7 @@ class Bad_Checks implements themecheck {
 			'/pub-[0-9]{16}/i' => __( 'Google advertising code detected', 'theme-check' ),
 			'/sharesale/i' => __( 'Sharesale affiliate link detected', 'theme-check' ),
 			'/affiliate_id/i' => __( 'Potential affiliate link detected', 'theme-check' ),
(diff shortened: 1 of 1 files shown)


pattonwebz:
I have done some very basic initial tests on this PR and it seems to function as expected.

acosmin:
@aaronjorbin If you could also add these ones :) It would be great:
https://wpdirectory.net/search/01EF4V4ATY7GV50PP5S712N7J2
https://wpdirectory.net/search/01EF4V5TAQFS4RDXTJ8XAAA64C

Thanks!

pattonwebz:
@acosmin I can't see a good way to match the 2nd search there as it seems like it can be used in help articles as well but the first one can be just an exact match. Will see about adding that and do a little more testing on this :)

pattonwebz:
@acosmin I managed to add a check that caught both. Can you test this to verify it works as expected?

acosmin:
@pattonwebz looks ok to me 👍

aaronjorbin:
That change looks good to me to

PR #272, opened 2020-08-15
add new allowable 'block-patterns'  and 'full-site-editing' tags

### LINKED ISSUE #271: Update allowed tags list
The allowed tags list in the checker is a static list which does not match what the filters show on .org.

Specifically the 'block-patters' tag is missing. It has not been noticed till now because the patterns were only introduced in WP 5.5 this last week.

pattonwebz:
Adds the 'block-patterns' tag as an [allowed tag for themes to use](https://make.wordpress.org/themes/handbook/review/required/theme-tags/#features-tags) when adding their theme to .org directory

Fixes: #271

### DIFF
--- a/checks/style_tags.php
+++ b/checks/style_tags.php
@@ -64,65 +64,65 @@ class Style_Tags implements themecheck {
 
 	function getError() { return $this->error; }
 
-    /**
-     * Get full list of allowed tags - including subject tags.
-     *
-     * @return array
-     */
+	/**
+	 * Get full list of allowed tags - including subject tags.
+	 *
+	 * @return array
+	 */
 	public static function get_allowed_tags() {
 		$allowed_tags = array(
-            'grid-layout',
-            'one-column',
-            'two-columns',
-            'three-columns',
-            'four-columns',
-            'left-sidebar',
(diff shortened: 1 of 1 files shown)


checks/style_tags.php
```diff
@@ -64,5 +63,67 @@ function check( $php_files, $css_files, $other_files ) {
 	}
 
 	function getError() { return $this->error; }
+
+	/**
+	 * Get full list of allowed tags - including subject tags.
+	 *
+	 * @return array
+	 */
+	public static function get_allowed_tags() {
```
dingo-d, line 72: Why `static`, is it being used somewhere else as a helper?

pattonwebz: It is used as a unitily helper function without any coupling to the class instance so static is what I generally give these types of methods. It is not being used anywhere else though so it really would have made no different if I made it public or private lol

dingo-d: Yeah, if it's used just in this class, private would make more sense :D

pattonwebz: Make these private methods instead :)

dingo-d (approved):

Issue #279, opened 2020-11-05
Adapt checks for Full-Site Editing themes


fklein-lu:
Full-site editing themes do not use mosts of the existing PHP template tags, CSS classes, or theme supports. They therefore currently do not pass the Theme Checker.

We should implement a way to modify the existing checks when an FSE theme is uploaded. These include:

- Required:
   - Tag display.
   - CSS classes
   - Avatar display
   - Hardcoded titles
   - Post navigation
   - Comment reply script
   - All basic checks
- Recommended:
   - Widget support
   - Navigation menu support
   - Post thumbnails support
   - Custom header and custom background support
   - Editor styles support

At the same time we should ensure that FSE themes have at least an `index.html` template, and a theme config file.

joyously:
Would there be a way to check if there are corresponding blocks in the templates for post navigation and tags and avatars? How does that other stuff happen (comment reply, widget support, post thumbnail support)?

fklein-lu:
The point of this issue is to enable developers _right now_ to upload FSE themes without needing to include code that they don't need just for the sake of passing the Theme Checker.

There are ways to check FSE themes for compliance, and I'm sure that will be done in the future.

But FSE is still experimental, and therefore there are no guidelines. So including checks in the Theme Checker before FSE is stable, and the guidelines have been updated is putting the cart before the horse.

Otto42:
The theme checker should be modified to add support for ways of checking for these things, certainly, but if these "FSE" themes are missing basic requirements, then they should not be added to the theme directory.

Disabling checks just to allow these is premature. Themes should be complete and cover all the use cases for them. If any "FSE" themes have used this approach to bypass requirements, then they should be removed from the theme directory. Merely being "new" does not absolve them from being complete as well.

aristath:
All themes go through a manual review once they get uploaded to the repo. Nothing goes live without first being reviewed, and this theme-check is just the 1st step in the process. It doesn't however stop there... This is just some sanity checks that run on theme-upload to avoid some of the most common errors, security issues and bad practices.
Right now block-based themes can't even be uploaded and go to the step where they can be manually reviewed, that's what needs to be fixed. If there are any issues the manual review will find them, but first the theme needs to be uploaded to the repo.

carolinan:
**Currently** the theme-json / experimental-theme.json is not required for FSE to work, only for global styles to work.

carolinan:
A first version has been added with https://github.com/WordPress/theme-check/pull/280.

[...]

PR #280, opened 2020-11-05
Adapt checks for Full-Site Editing themes

fklein-lu:
This code is "good enough", by design. It's supposed to make it easier to upload experimental themes to WordPress.org without having to add code just for the sake of passing the checker.

As FSE themes become more stable, the Theme Checker as well as the checks would need a larger re-organization. So this can be revisited as part of that change down the line.

See #279.


dd32:
While "good enough" I'm pondering if maybe each check should just have a flag added to it which designates what kind of theme it's for, and have the runner disregard non-FSE checks for FSE themes and vice versa..

I agree that that could be changed down the line though.

carolinan:
After discussions with the theme's team representatives, it was decided that we should move forward with this version and iterate.

PR #281, opened 2020-11-05
Fix false positives in title checks

fklein-lu:
When `strpos` cannot find a string, it will return `false`. Therefore all these checks need to compare against `false`, else this leads to false positives.

To test, use a theme that has neither `title-tag` support, nor `<title>` or `wp_title()` anywhere. You'll see that the checker will nonetheless throw these errors:
```
REQUIRED: The theme must not use the <title> tags. Found the tag in index.php.
REQUIRED: The theme must not use the <title> tags. Found the tag in inc/class-require-gutenberg.php.
REQUIRED: The theme must not use the <title> tags. Found the tag in functions.php.
REQUIRED: The theme must not call to wp_title(). Found wp_title() in index.php.
REQUIRED: The theme must not call to wp_title(). Found wp_title() in inc/class-require-gutenberg.php.
REQUIRED: The theme must not call to wp_title(). Found wp_title() in functions.php.
```

### DIFF
--- a/checks/title.php
+++ b/checks/title.php
@@ -28,7 +28,7 @@ class Title_Checks implements themecheck {
 
 			// Look for <title> and </title> tags.
 			checkcount();
-			if ( ( 0 <= strpos( $php, '<title>' ) || 0 <= strpos( $php, '</title>' ) ) && ! $titletag  ) {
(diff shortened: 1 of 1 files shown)


dd32 (approved):
Using `false !== strpos()` here would be correct, switching to the PHP8 `str_contains()` would also be an option with a polyfill loaded for <PHP8.

Issue #200, opened 2019-01-26
False positive for add_theme_page


danyj:
add_*_page  https://github.com/WordPress/theme-check/blob/da79a6f17b126b52998474cd234e4210de5a2d6f/checks/admin_menu.php#L34

sees this https://github.com/Themezly/Creatus/commit/1cc280519583bec28c124b6f348472ecdae114c2#diff-3f431296b9b6636a53b516cf26b3b1fbL7576

as `add_theme_page`. That regex should check that there are no spaces in between.

PR #284, opened 2020-11-19
Fix false positive for add_theme_page

carolinan:
Fixes https://github.com/WordPress/theme-check/issues/200 by checking that there is no white space.

How to test:
Add the following to any theme .php file:

```
function wpdocs_plugin_menu() {
    add_dashboard_page( __( 'WPDocs Plugin Dashboard', 'textdomain' ), __( 'WPDocs Plugin', 'textdomain' ), 'read', 'wpdocs-unique-identifier', 'wpdocs_plugin_function' );
}
add_action('admin_menu', 'wpdocs_plugin_menu');

$add_css = thz_page_block_positions_css();

function thz_page_block_positions_css() {

};

```

The result should be a warning for add_dashboard_page, but not for `add_css = thz_page()`


checks/admin_menu.php
```diff
@@ -31,7 +31,7 @@ function check( $php_files, $css_files, $other_files) {
 // Note to TGMPA: Stop trying to bypass theme check. 
 
 		$checks = array(
-			'/(?<!function)[^_>:](add_[^_\'",();]+?_page)/' => _x( 
+			'/(?<!function)[^_>:](add_[^_\'",();][^\s]+?_page)/' => _x(
```
dd32, line 34: Could this just be simplified to.. ?
```suggestion
			'/(?<!function)[^_>:](add_[a-z]+_page)/' => _x( 
```

carolinan: This code is so old I don't think anyone would remember the reason to why it was originally written this way.

Issue #149, opened 2016-04-27
Add a check for the <script> tag in the header.php


grappler:
All scripts need to be enqueued. The exception is for this at the moment loading browser specific scripts in a function. We can soon remove the exception soon as `wp_script_add_data()` was introduced in WP 4.1

Otto42:
The wp_add_inline_script was only added in WordPress 4.5. This will need to be a RECOMMENDED for a few versions before it can be a requirement.

grappler:
We are already requiring themes to add inline script using `wp_head`. This check would only check the header.php for themes adding scripts like html5.js and similar. We could theoretically see if the tag contains `src` but themes should not be adding scripts in the header as it becomes a bit more difficult to remove. It is easier to remove with `remove_action`

carolinan:
This is still a very common problem.

Here are some examples, but they also include styles and footer.php:

https://themes.svn.wordpress.org/photography-simple/1.2/header.php
https://themes.svn.wordpress.org/blogos/0.3/header.php
https://themes.svn.wordpress.org/montero/1.0.0/header.php
https://themes.svn.wordpress.org/midium/1.0/footer.php
https://themes.svn.wordpress.org/ilse-marie/1.0.3/footer.php

PR #285, opened 2020-11-19
Add a check for the <script> tag

carolinan:
Fixes https://github.com/WordPress/theme-check/issues/149

It does not stop upload since there can be valid exceptions that needs manual review.

### DIFF
--- a/checks/script_tag.php
+++ b/checks/script_tag.php
@@ -23,7 +23,14 @@ class Script_Tag implements themecheck {
 			if ( preg_match( '/<script/i', $file_content ) ) {
 				$error         = '/<script/i';
 				$grep          = tc_preg( $error, $file_path );
-				$this->error[] = sprintf( '<span class="tc-lead tc-required">' . __( 'REQUIRED', 'theme-check' ) . '</span>: ' . __( 'Found a script tag in %s. Scripts and styles needs to be enqueued or added via a hook, not hard coded.', 'theme-check' ),
+				$this->error[] = sprintf(
+					'<span class="tc-lead tc-required">' . __( 'REQUIRED', 'theme-check' ) . '</span>: %s %s',
+					sprintf(
+						__( 'Found a script tag in %s. Scripts and styles needs to be enqueued or added via a hook, not hard coded.', 'theme-check' ),
+						'<strong>' . $filename . '</strong>'
+					),
+					$grep
+				);
 				'<strong>' . $filename . '</strong>' ) . $grep;
 			}
 		}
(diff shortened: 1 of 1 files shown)


checks/script_tag.php
```diff
@@ -0,0 +1,32 @@
...
+
+		$ret = true;
+
+		checkcount();
+		/**
+		 * This check is limited to header.php and footer.php.
+		 */
+		foreach ( $php_files as $file_path => $file_content ) {
+			$filename = tc_filename( $file_path );
+			if ( 'footer.php' === $filename && preg_match( '/<script/', $file_content ) || 'header.php' === $filename && preg_match( '/<script/', $file_content )) {
```
dd32, line 19:
```suggestion
			if ( ! in_array( $filename, array( 'header.php', 'footer.php' ) ) ) {
				continue;
			}

			if ( preg_match( '/<script/i', $file_content ) ) {
```

Seems much easier to read in this case, and checking for `<script` case insensitively.

checks/script_tag.php
```diff
@@ -0,0 +1,32 @@
...
+				$error         = '/<script/';
+				$grep          = tc_preg( $error, $file_path );
+				$this->error[] = sprintf( '<span class="tc-lead tc-required">' . __( 'REQUIRED', 'theme-check' ) . '</span>: ' . __( 'Found a script tag in %s. Scripts and styles needs to be enqueued or added via a hook, not hard coded.', 'theme-check' ),
```
dd32, line 22: A common thing I'm seeing in Theme Check is that there's often a lack of spaces between error and the examples. This seems simpler and easier to read:

```suggestion
				$this->error[] = sprintf(
					'<span class="tc-lead tc-required">' . __( 'REQUIRED', 'theme-check' ) . '</span>: %s %s',
					sprintf(
						__( 'Found a script tag in %s. Scripts and styles needs to be enqueued or added via a hook, not hard coded.', 'theme-check' ),
						'<strong>' . $filename . '</strong>'
					),
					$grep
				);
```

checks/script_tag.php
```diff
@@ -0,0 +1,32 @@
...
+				'<strong>' . $filename . '</strong>' ) . $grep;
```
dd32, line 23:
```suggestion
```

dd32 (changes requested):
Just some thoughts on making the check read a bit simpler, and error formatting.

checks/script_tag.php
```diff
@@ -0,0 +1,36 @@
...
+		 * This check is limited to header.php and footer.php.
+		 */
+		foreach ( $php_files as $file_path => $file_content ) {
+			$filename = tc_filename( $file_path );
+			if ( ! in_array( $filename, array( 'header.php', 'footer.php' ) ) ) {
+				continue;
+			}
+
+			if ( preg_match( '/<script/i', $file_content ) ) {
+				$error         = '/<script/';
```
carolinan, line 24:
```suggestion
				$error         = '/<script/i';
```

Issue #286, opened 2020-11-29
fwrite (composer)


benlumia007:
I'm trying to upload a theme update but for some reason, it failed but 

```
WARNING: More than one text-domain is being used in this theme. This means the theme will not be compatible with WordPress.org language packs.The domains found are camaraderie, backdrop-core
WARNING: fwrite was found in the file vendor/composer/platform_check.php File operations should use the WP_Filesystem methods instead of direct PHP filesystem calls.Line 17: fwrite(STDERR, 'Composer detected issues in your platform:' . PHP_EOL.PHP_E
RECOMMENDED: No reference to add_theme_support( "custom-background", $args ) was found in the theme. If the theme uses background images or solid colors for the background, then it is recommended that the theme implement this functionality.
RECOMMENDED: Theme URI: is missing from your style.css header.
RECOMMENDED: Author URI: is missing from your style.css header.
```
 i didn't had any issues before when i updated a theme update on 11.15.2020

any suggestions or idea

benlumia007:
or could this been somethign to do with a platform_check for composer 2.0

benlumia007:
i found out that is composer 2.0.5 or higher that has the new fwrite warning. so i downgraded to Composer 2.0.4. Should this be fixed in theme check.

dd32:
As this is only a warning, and `fwrite()` is discouraged from theme usage, this warning is fine.

Manual review would easily see that it's inside composer and unrelated to the theme, unfortunately the `fwrite()` warning is currently marked as a critical failure and blocks upload, which will be fixed via #287

If you want to skip including the platform check code that Composer generates, you can check their docs: https://getcomposer.org/doc/07-runtime.md#platform-check

benlumia007:
@dd32 
Thanks for the explanation

PR #287, opened 2020-12-03
Allow file_get_contents() and discourage usage of WP_Filesystem for basic IO tasks

dd32:
This PR removes warnings about `file_get_contents()` and encourages it's usage over WP_Filesystem for general IO operations.

HTTP methods are also replaced with a new error message, to use the WordPress HTTP library instead.

### DIFF
--- a/checks/filesystem-http.php
+++ b/checks/filesystem-http.php
@@ -36,7 +36,6 @@ class FilesystemHttpCheck implements themecheck {
 							$check,
 							$grep
 						);
-						$ret = false;
 					}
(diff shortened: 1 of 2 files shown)


aristath (approved):
I love this. Looks good :+1: :heart:

benlumia007:
I just wanted to make sure, since I open the #286 ticket about the fwrite and it has been a while.  I'm still using the older version. is it safe to upgrade to the latest version of composer when trying to upload themes to the repo.

PR #289, opened 2020-12-03
Add a deploy action, to push it to the WordPress.org plugin directory

dd32:
The attached GitHub workflow deploys the plugin to the plugin directory using [10up's plugin deploy action](https://github.com/10up/action-wordpress-plugin-deploy).

It's triggered by a Release being created on GitHub.

To release a new version with this:
 - update plugin header version string
 - update readme.txt stable tag
 - tag/release a release
 - wait a few minutes and see it's updated on WordPress.org.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,6 +6,10 @@
 /.github export-ignore
 /.phpcs.xml.dist export-ignore
 /.stylelintrc export-ignore
+/.gitignore export-ignore
+/.gitattributes export-ignore
+/composer.json export-ignore
+/composer.lock export-ignore
 
 #
 # Auto detect text files and perform LF normalization
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,28 @@
+name: Deploy to WordPress.org
+on:
+  release:
+    types: [published]
+jobs:
+  tag:
+    name: New release
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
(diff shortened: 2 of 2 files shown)


dingo-d:
Would it be a good idea to add `.distignore` file with all the things you'd want to exclude from the release of the plugin?

For instance `composer.json`, `composer.lock`, `.gitignore`, `.gitattributes`, `phpcs.xml.dist` and the `.github` folder could all be excluded.

dd32:
> Would it be a good idea to add `.distignore` file with all the things to exclude from the release of the plugin?

`.gitattributes` already takes care of some of them I believe
 - `.github`
 - `.phpcs.xml.dist`
 - `.stylelintrc`

That leaves
 - `composer.json`
 - `composer.lock`
 - `.gitignore`
 - `.gitattributes`

Looking at https://github.com/10up/action-wordpress-plugin-deploy/blob/develop/entrypoint.sh the above should indeed be added to our `.gitattributes` I think.

dingo-d:
I had some issues with `.gitattributes` when I played around with this action, I think it's safer to go with `.distignore`. I mean, if it works with `.gitattributes` then great 😁

dd32:
https://github.com/WordPress/theme-check/pull/289/commits/04de6a33c26fcdeab07efeb5ccd64c6a8caf5bdc `.gitattributes` seems to have worked to exclude them without issue for me.

Unless there's a reason for including them in GitHub releases without also including them dotorg releases I don't think it's a good thing to duplicate it..

dingo-d (approved):

PR #291, opened 2020-12-03
Avoid false positives when .svn/.git/.hg/.bzr exists within the path to a theme.

### LINKED ISSUE #193: False positive result returned from directories.php
I've been frustratingly trying to diagnose the error "Please remove any extraneous directories like .git or .svn from the ZIP file before uploading it."

When digging into the source code, the $name variable contains the full path to the file and is ran through strpos to detect the presence of .git, .svn, .hg and .bzr. This throws a false positive though because our domain is www.hg...

`foreach ( $css_files as $name => $file ) {
			checkcount();
			if ( strpos( $name, '.git' ) !== false || strpos( $name, '.svn' ) !== false || strpos( $name, '.hg' ) !== false || strpos( $name, '.bzr' ) !== false ) $found = true;
		}`

I confirmed this is the case by removing the strpos function for .hg and the theme validated successfully. It would be good if the `$name` string could be split to only check inside the theme folders.

dd32:
Combines the logic and avoid false positives where .svn/.git/.hg/.bzr exist within a filename/file path but are not actually present in the theme.

An example of this would be having Theme Check installed in a WordPress install at `~/public_html/company.hg/`.

This also extends the `.hg` and `.bzr` checks to PHP files.

Fixes #193

### DIFF
--- a/checks/directories.php
+++ b/checks/directories.php
@@ -5,27 +5,30 @@ class DirectoriesCheck implements themecheck {
 
 	function check( $php_files, $css_files, $other_files ) {
 
+		$excluded_directories = array(
+			'.git',
+			'.svn',
+			'.hg',
+			'.bzr',
+		);
+
 		$ret = true;
-		$found = false;
 
-		foreach ( $php_files as $name => $file ) {
+		$all_filenames = array_merge(
+			array_keys( $php_files ),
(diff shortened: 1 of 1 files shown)


carolinan:
LGTM, found no problems or false positives in my test.

PR #294, opened 2020-12-07
Introduce run_themechecks_against_theme() to allow running against a WP_Theme instance

### LINKED ISSUE #160: Allow Text Domain check to work on w.org
The text domain was failing on w.org because of the way the Theme Check is being loaded.

[`WPORG_Themes_Upload::check_theme()`](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-upload.php#L468) directly calls `checkbase.php` without loading `main.php` which defines data about the theme which is need for the text domain check.

dd32:
This PR aims to fix #160 by introducing a generic function that allows it to be run against a WP_Theme instance, rather than relying upon global state.

This resulted in a fair bit of cleanup and reliance upon WP_Theme, which reduces the code complexity.

This also required extending the `themecheck` interface (that should've been a class) with a method to pass extra context to the theme checks, it's not present in the interface as it would be then required to be defined by any custom theme checks that exist in the wild. Adding an extra parameter to `check()` is also not possible for the same reason.

The diff of this PR is rather large due to some badly nested branches that have been removed or rendered not needed.


carolinan:
Happy to see global removed, reviewed to the best of my ability, no issues found.

If anything, should we prepare for separating html files. not just php, css and "other" ?

carolinan:
What I meant was, should html files be separated to prepare for future checks for FSE HTML templates?
Or do you see that as being checked by a different tool that needs to be built separately.

dd32:
I see no urgent need to add splitting out for HTML files yet, once a check is needing to be added that needs that, we can do it at that point in time.
Right now, I have no idea what is needed for FSE other than #280 

(I'll be circling back to my in-progress theme check stuff soon)

PR #320, opened 2021-06-01
Remove block-patterns recommendations

aristath:
Block-patterns are not a requirement. It may be desirable for a theme to have block-patterns, but that is rather opinionated and not all themes should have patterns. A pattern is basically just reusable content and has absolutely nothing to do with a theme. This should not exist in the theme-check plugin.


carolinan:
I would prefer to have it changed to info in that case because many authors don't know that these are allowed.

aristath:
Is the role of theme-check to inform theme-authors that they can do something? :thinking:

carolinan:
The plugin reaches more theme authors than themes team can alone. So yes, when there are new features that we want to establish.

aristath:
Closing this one after discussing further.

PR #321, opened 2021-06-01
Remove content_width recommendations

aristath:
The content_width is no longer important in core. It is only used for embeds widths, and even there it is not that important because embeds in WP have been responsive for a long time.
I think we can remove this check completely...


kafleg:
Do this need to check it? @aristath

aristath:
yes, this needs to be checked :+1:

PR #333, opened 2021-06-08
Fix wp_get_theme data is not fetched

### LINKED ISSUE #332: wp_get_theme data is not fetched
In the area where the theme information is presented visually, and in checks that uses theme data like theme author name, license and URI's, not all theme data is fetched and instead, returns null.

Expected result:
image

Actual result:
image

carolinan:
Fixes #332

Changes `$theme['item']` to `$theme->get( 'item' )`

#### Theme Info:
1. Apply the PR
2. In the plugin, select a parent theme to check and confirm that all theme details are showing.
3. Select a child theme to check and confirm that all details are showing.

#### URI Check:
1. In a non-default theme, change theme and author URI to https://wordpress.org/themes/twentytwentyone/
2. Check the theme with the plugin. Confirm that there are two errors:
```
REQUIRED: Your Theme URI and Author URI must not be the same.
REQUIRED: Using a WordPress.org Theme URI is reserved for default and bundled themes (Twenty * series).
```
3. Select and check a default theme.
4. Confirm that there is no error saying that the Theme URI can not be wordpress.org.

#### Link_Check (hard coded links):
1. In your theme,  add `<a href="https://example.com">example</a>` to any .PHP file.
2. Check the theme with the plugin and confirm that the info is showing
`INFO: Possible hard-coded links were found in the file filename. Line number: <a href='https://example.com'>example</a>`

#### Underscores check:
1. In your theme, replace theme and and author URI with underscores.me.
2. Check the theme with the plugin and confirm that the error is showing:
`REQUIRED: Found a copy of Underscores. Using underscores.me as Theme URI or Author URI is not allowed. Update the files for your own theme.`


carolinan:
_REQUIRED: Your Theme URI and Author URI must not be the same._
This needs to be reduced from required, should not block upload.

PR #346, opened 2021-06-16
Reduce the favicon check from required to info

kafleg:


carolinan (changes requested):
"check" needs to return true otherwise the upload is blocked.

PR #348, opened 2021-06-17
Update changelog.txt

carolinan:


dd32:
Does readme.txt also need updating?

carolinan:
No I don't think that is needed.

kafleg:
What is the importance of mentioning Changelog in readme.txt if we have a separate file for changelog?

PR #354, opened 2021-06-18
Fixed main.php

kafleg:
Offset Undefined offset: 1 in theme-check\main.php on line 99

### DIFF
--- a/main.php
+++ b/main.php
@@ -96,7 +96,7 @@ function check_main( $theme_slug ) {
 		'<strong>' . esc_html( $checkcount ) . '</strong>',
 		'<strong>' . esc_html( $theme['Title'] ) . '</strong>',
 		'<strong>' . esc_html( $version[0] ) . '</strong>',
-		'<strong>' . esc_html( $version[1] ) . '</strong>'
+		'<strong>' . esc_html( $version[0] ) . '</strong>'
 	) . '</p>';
 
 	$results = display_themechecks();


carolinan:
I am not able to reproduce the problem. How is the error triggered? Which PHP version is used?

kafleg:
> I am not able to reproduce the problem. How is the error triggered? Which PHP version is used?

PHP version 7.4 used. 

While checking themes, I can see it. 
image

carolinan:
I have been able to update my PHP version and reproduce the PHP problem, but the PR does not solve the issue.

Please look through the code again and consider what it does.
It displays the version of the guidelines, and the plugin revision number. If you look at the screenshot above,
there is no number output after the text "plugin revision"

`$plugins = get_plugins( '/theme-check' ); ` -First, fetch the information from the plugin header.
See https://developer.wordpress.org/reference/functions/get_plugins/

`$version = explode( '.', $plugins['theme-check.php']['Version'] );`
Next, explode the version at the `.`.
-This fails because when then plugin was updated, the revision was not added to the version number, there is no `.`

Hence, the `$version` array only contains 0, not 1: `Array ( [0] => 20210617 )` and it can not print `$version[1]`

When you change the  `$version[1]` to  `$version[0]` you are asking it to print  `$version[0]` twice, and output is incorrect.

carolinan:
My suggestion would be to first solve the problem with the title check that is blocking upload, then release a new version with a revision number added.

To avoid future PHP issues this file could updated to first check if the array keys exist before printing the text.

PR #360, opened 2021-07-01
Update allowed tags in the report

carolinan:
Update allowed tags in the report so that code examples and links work again.

Before:
Report without working links

After:
Report displays the code examples and links with improved readability.

### DIFF
--- a/main.php
+++ b/main.php
@@ -135,6 +135,11 @@ function check_main( $theme_slug ) {
 				'class' => array(),
 			),
 			'strong' => array(),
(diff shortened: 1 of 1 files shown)


kafleg (commented):
Errors are formatting in a proper way as shown in the above screenshot. But the closing bracket in some code lines is not proper. 
image

Links are working fine. But i suggest to open the link in new tab.

carolinan:
The spacing of the code actually looks that way in the theme, they are mixing tabs and spaces.

Issue #356, opened 2021-06-22
Functions is being misread as a default WP function on theme update


eramits:
Hello,

We have a little problem with the new update checker rules. One of our functions (oceanwp_title()) is being misread as a default WP function (wp_title(), which is not in use in the theme at all), and we're unable to push an update. oceanwp_title has been in the theme since the very first version. We could change the function name, but first, we wanted to check if there's a way to bypass this, otherwise, all our users using custom codes related to this function will experience issues. Not to mention other possible and future issues with all functions using oceanwp in the name. Grateful for any help and feedback we can get 

https://github.com/WordPress/theme-check/blob/bba5fefad5ae10f1cf50070e9260230770039d9f/checks/class-title-check.php#L38

<img width="819" alt="dsdsd" src="https://user-images.githubusercontent.com/43537450/122873401-5660e580-d34f-11eb-82f9-0d3ea17867c1.PNG">

Thank you

aristath:
The problem here is that we check if `wp_title` is contained, and `oceanwp_title` contains `wp_title`.
Ideally we'd look for exact word matches, so using a regex function that would be something like `\bwp_title\b`
Maybe we should introduce a function `themecheck_function_located( $function_name )` and use that consistently everywhere instead of the `strpos` calls? :thinking: 
cc @carolinan

carolinan:
Yes, a regex would be an improvement. I question whether this check should block upload at all.

carolinan:
I found 11 themes (including oceanwp) that were blocked by this requirement since june 17.

The requirement is in place because wp_title was supposed to be deprecated many years ago, and it was reverted because so many themes still used it incorrectly. `add_theme_support( 'title-tag' );` replaced it.

aristath:
Should the check be removed, or tweaked?
Tweaking it should be relatively easy and changing the line highlighted above to this should do it:
```php
if ( preg_match( '/\bwp_title\b/', $file_content ) ) { 
```

carolinan:
tweaked then kept as required :)
The message needs to be more clear and explain why.

PR #362, opened 2021-07-01
Reduce the title tag and wp_title checks to recommended

carolinan:
This PR reduces the title tag and wp_title checks to recommended.
Updates the message to explain why it is not recommended to use them, and what to use instead.
Adds line numbers to the message.

Partial for https://github.com/WordPress/theme-check/issues/356
It should solve the issue, but does not address the suggestion to add a new function for general use.


kafleg (commented):
Tested by adding wp_title() in theme code.
The RECOMMENDED message is fine.

carolinan:
I will merge this, -if there are false positives for the wp_title, it no longer blocks upload.
Thank you for testing.

Issue #341, opened 2021-06-15
Mention only major version for Requires PHP and Tested up to


kafleg:
In theme style.css file, it should be mentioned
```
Requires PHP: 5.2
Tested up to: 5.7
```
but most of the theme author mentioned minor version as well. For example,
```
Requires PHP: 5.2.3
Tested up to: 5.7.2
```

As @Otto42 said that it is only recommended to mention the major version. His [comment](https://themes.trac.wordpress.org/ticket/84095#comment:2) in a ticket,

> While we are not using the "Tested up to" parameter yet for themes, for the case of plugins, we generally recommend leaving it at the major version number only. That is, WordPress is at 5.4.1, but Tested up to can be at 5.4 without any issues.
> 
> In the plugin directory, we automatically assume that minor versions are compatible if you specify it in this manner, and will show the latest version without you needing to update to every minor version.
> 
> When this gets implemented for themes, I expect that we will use the same basic formula. So "Tested up to: 5.4" is fine. No need to specify 5.4.0 or 5.4.1 and so on. Just for the future.
>

TeBenachi:
#373

carolinan:
I am closing the issue since this has been completed. 🎉

PR #364, opened 2021-07-02
Check requires php versions

TeBenachi:


TeBenachi:
#341

I’m also trying to create a check for "Tested up to" but having a trouble figuring out. 
What is the value used  for “Tested up to”? 

`$this->theme->get( 'Testedupto' )`

I tried `Testedupto` but it did not work.  Can anyone help please?

carolinan:
Hi -I had to look it up but it turns out, _Tested up to_ is not part of WP_Theme class:
https://developer.wordpress.org/reference/classes/wp_theme/#source

In the check that tests if the _Tested up to_ exists, the CSS file is parsed:
https://github.com/WordPress/theme-check/blob/master/checks/class-style-css-header-check.php

Please submit one PR per issue, I mean that the plugin version update should be in it's own pull request.
(The version would be updated both in the main plugin file and the readme file, and the changelog updated.)

checks/class-version-php-check.php
```diff
@@ -0,0 +1,73 @@
+<?php
+/**
+ * Check for wp_title and title tags.
```
carolinan, line 3: Please update all the code comments to match what the check and the code does.

checks/class-version-php-check.php
```diff
@@ -0,0 +1,73 @@
...
+	 * Check that return true for good/okay/acceptable, false for bad/not-okay/unacceptable.
+	 *
+	 * @param array $php_files File paths and content for PHP files.
+	 * @param array $css_files File paths and content for CSS files.
+	 * @param array $other_files Folder names, file paths and content for other files.
+	 */
+	public function check( $php_files, $css_files, $other_files ) {
+
+		checkcount();
+		$ret = true;
```
carolinan, line 46: Because this is a recommendation, `$ret = true;` and later` $ret  = false;` are not needed.
`$ret  = false;` would prevent themes from being uploaded to the theme directory.
Instead, the `check` function should return true.

TeBenachi:
Hi @carolinan   Thanks so much for your insights! 
I appreciate you always taking time to share very helpful resources!

carolinan:
And thank you for working on this.
The check works well. The code standard issues need to be fixed before the pull request can be merged.
The version change needs to be reverted. We can update that when we have completed some more fixed and the plugin can be released.

TeBenachi:
Thank you for checking!  
I just submitted the Tested up to check. But I realized the code standard issues need to be fixed and the version needs to be reverted back as you mentioned above.

PR #373, opened 2021-07-13
Add version check for tested up to

TeBenachi:


carolinan:
This is super minor 😄 but do you think would make sense to use a version number that matches WordPress?, I think this example version number 7.4.1 was copied from the `Requires PHP` check?

TeBenachi:
Yep I copied 😁 I think it’s less confusing to the authors too. Thanks for checking!

carolinan:
Thank you.

I'm not sure how we missed this,
`$file_path  = get_stylesheet_directory() . '/style.css';`
Looks at the stylesheet of the active theme, not the theme that is selected in the list in the plugin.

TeBenachi:
@carolinan Great catch! Thanks for noticing. 
The path to the stylesheet for the selected theme is corrected.

checks/class-version-tested-check.php
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Check for "Tested up to" versioning
+ *
+ * @package Theme Check
+ */
+
+/**
+ * Check for "Tested up to" versioning.
+ * Does "Tested up to" include patch versions (e.g. 5.8.1)?
+ * If so, recommend including major and minor verisions only (e.g. 5/8)
```
carolinan, line 11:
```suggestion
 * If so, recommend including major versions only (e.g. 5.8)
```

checks/class-version-tested-check.php
```diff
@@ -0,0 +1,76 @@
...
+			)
+		);
+
+		if ( ! empty( $theme_data['TestedUpto'] ) ) {
+			$req_tested_decimal_count = substr_count( $theme_data['TestedUpto'], '.' );
+			if ( $req_tested_decimal_count > 1 ) {
+				$this->error[] = sprintf(
+					'<span class="tc-lead tc-recommended">%s</span>: %s',
+					__( 'RECOMMENDED', 'theme-check' ),
+					__( '<strong>Tested up to</strong> is recommended to have major and minor versions only (e.g. 5.8). Patch version is not needed (e.g. 5.8.1).', 'theme-check' )
```
carolinan, line 59:
```suggestion
					__( '<strong>Tested up to</strong> is recommended to have major versions only (e.g. 5.8). Patch version is not needed (e.g. 5.8.1).', 'theme-check' )
```

carolinan: Confirmed with https://make.wordpress.org/core/handbook/about/release-cycle/version-numbering/
5.8 is a major, 5.8.1 is minor.

carolinan:
> @carolinan Great catch! Thanks for noticing.
> The path to the stylesheet for the selected theme is corrected.

It works correctly in my testing.

TeBenachi:
Great! Thanks!!

TeBenachi:
🎉🎉🎉

PR #377, opened 2021-07-30
Added Template Editing tag in allowed tags list

kafleg:
Template Editing is a new tag and added in WordPress.  Current version of Theme checks show it as REQUIRED.


kafleg:
image

Issue #382, opened 2021-08-12
Check if the theme has a copyright notice


carolinan:
GPL **requires** that there is a copyright notice for the theme.
The easiest description of this is in the howto: https://www.gnu.org/licenses/gpl-howto.html

Example formats from gnu: `Copyright 1998, 1999 Terry Jones`, `Copyright 1998-2021 Terry Jones`.

First, we know that themes do not use consistent formatting or placement for the copyright notice.
Secondly, third party assets includes a copyright notice (as they should) that the check needs to ignore.
And third, theme authors often copyrights their work to:
- WordPress.org, WordPress, WordPress.org team (they copy default themes)
- Automattic (they copy underscores)
- Fred 🤷‍♀️

because Fred is used in the example:
https://make.wordpress.org/themes/2014/07/08/proper-copyrightlicense-attribution-for-themes/

Finally, the copyright notice **should not** be included if the theme license is public domain, which happens but is rare.

I think the best option is to add a check that requires the word "Copyright" to be inside either style.css or readme.txt,
but not check for further formatting.

kafleg:
That word might be the best option to check on those files.  Some themes can have a © symbol too. We can add an INFO about this.

carolinan:
The current check is very limited, but better than none, and a reminder to theme authors.

PR #388, opened 2021-08-18
Create class-copyright-notice-check.php

carolinan:
Partial fix for  https://github.com/WordPress/theme-check/issues/382

_This check is pretty useless_. It does not cover the actual theme directory requirement.
It checks readme.txt and readme.md and CSS files for `Copyright`, `copyright`, and `(C)`.
But it can't tell if the mention of copyright is for the theme itself or an asset.

### DIFF
--- a/checks/class-copyright-notice-check.php
+++ b/checks/class-copyright-notice-check.php
@@ -1,0 +1,66 @@
+<?php
+/**
+ * Checks if the theme has a copyright notice.
+ *
+ * @package Theme Check
+ */
+
+/**
+ * Checks if the theme has a copyright notice.
+ */
+class Copyright_Notice_Check implements themecheck {
+	/**
+	 * Error messages, warnings and info notices.
+	 *
+	 * @var array $error
+	 */
+	protected $error = array();
+
+	/**
+	 * Check that return true for good/okay/acceptable, false for bad/not-okay/unacceptable.
+	 *
+	 * @param array $php_files File paths and content for PHP files.
+	 * @param array $css_files File paths and content for CSS files.
+	 * @param array $other_files Folder names, file paths and content for other files.
+	 */
+	public function check( $php_files, $css_files, $other_files ) {
+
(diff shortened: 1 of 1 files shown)


TeBenachi:
I can confirm the check works when the style.css or readme.txt. don't have any mention of Copyright, copyright, or (c).

screenshot1

kafleg:
> Partial fix for #382
> 
> _This check is pretty useless_. It does not cover the actual theme directory requirement, which is to include a copyright notice for the theme itself.
> It checks readme.txt and readme.md and CSS files for `Copyright`, `copyright`, and `(C)`.
> But it can't tell if the mention of copyright is for the theme itself or an asset.

Yes. Copyright text is also found in footer copyright text, libraries, etc. So, this check work only for such theme where `Copyright`, `copyright`, and `(C)` is not mentioned. But, this is a very rare case I guess.

carolinan:
@kafleg It won't find copyright texts in the footer or libraries. 
It only checks style.css, readme.txt and readme.md

carolinan:
Setting this to draft again to remove some unused code and try to replace tc_filename().

carolinan:
@kafleg Now it only checks the content of:
themeslug/readme.txt
themeslug/readme.md
themeslug/style.css
themeslug/licence.txt

So it won't check sub folders.

kafleg:
> @kafleg Now it only checks the content of:
> themeslug/readme.txt
> themeslug/readme.md
> themeslug/style.css
> themeslug/licence.txt
> 
> So it won't check sub folders.
> It also shouldn't time out anymore.

Worked perfectly with all the above-mentioned files and in that location. Checked with the subfolder too. 

I just want to recommend, can't put WordPress.Org links(if have) for _Learn how to add a copyright notice (opens in a new window)._ instead of using a third-party site? 

P.S. I have no idea if that proper information page exists or not.

carolinan:
I think it is better to link to the actual source.

PR #397, opened 2021-09-07
Split the current theme path handling out of tc_filename() and optimize it to query once per theme

dd32:
As a follow up to #370, the change causes it to query all the files for a theme on every call to `tc_filename()` which on some themes can be a significant number.

This splits the function into two, and uses a static cache to only query the files once per theme tested.


carolinan:
I have tested with different types of themes and with #389 and I don't see any issues, timeouts or other.

PR #401, opened 2021-10-01
Update screenshot ratio from required to info

kafleg:


carolinan:
@kafleg What is the reason for this change?

PR #421, opened 2022-03-16
check for plugin territory of registerBlockType

kafleg:
Add registerBlockType in plugin territory. This PR is created based on the comment of Justin. 
https://themes.trac.wordpress.org/ticket/114545


kafleg:
Thanks for reviewing. @justintadlock :)

Issue #402, opened 2021-10-06
Warn about jQuery usage on the front end


adamsilverstein:
While it has been widely known that [you might not need jQuery](https://youmightnotneedjquery.com/) for most of your interactive theme features, jQuery is still a common JavaScript dependency found in WordPress themes - for example, 7 out of the 11 default themes since Twenty Ten use jQuery in the frontend. 

A recent analysis of [the performance impact of using jQuery in WordPress themes](https://make.wordpress.org/themes/2021/10/04/the-performance-impact-of-using-jquery-in-wordpress-themes/
) determined that **jQuery is the most common JavaScript-based performance problem in themes** and that **Removing jQuery may lead to 80% less JavaScript load**. According to the analysis, most themes can easily substitute (modern) vanilla JavaScript and remove their reliance on jQuery with only minor changes.

To help theme developers make better choices in the future, I would like to add a warning to the theme-check plugin when themes enqueue 'jquery' on the front end (eg. `wp_enqueue_script`). The warning could reference the analysis and perhaps offer a link to an [article explaining how to switch to vanilla JavaScript](https://tobiasahlin.com/blog/move-from-jquery-to-vanilla-javascript/). 

If this idea seems reasonable, I am happy to work on a PR to add this.

carolinan:
Yes,  I had it on my to do list but if you want to work on it that would be great.

There is also the scenario where the user tries to include and enqueue their own version of jQuery -that should block theme upload.

adamsilverstein:
@carolinan finally found time to get back to this and opened https://github.com/WordPress/theme-check/pull/423 to address.

I tested this locally to verify it works :) Still have some questions about the implementation which I will leave as comments on the PR code

PR #423, opened 2022-05-06
Add a front end JQuery enqueue check

adamsilverstein:
Addresses https://github.com/WordPress/theme-check/issues/402

### DIFF
--- a/checks/class-jquery-check.php
+++ b/checks/class-jquery-check.php
@@ -1,0 +1,55 @@
+<?php
+/**
+ * Check if JQuery is enqueued on the front end.
+ *
+ * @package Theme Check
+ */
+
+/**
+ * Check if JQuery is enqueued on the front end.
+ */
+class JQuery_Check implements themecheck {
+	/**
+	 * Error messages, warnings and info notices.
+	 *
+	 * @var array $error
+	 */
+	protected $error = array();
+
+	/**
+	 * Check that return true for good/okay/acceptable, false for bad/not-okay/unacceptable.
+	 *
+	 * @param array $php_files File paths and content for PHP files.
+	 * @param array $css_files File paths and content for CSS files.
+	 * @param array $other_files Folder names, file paths and content for other files.
+	 */
+	public function check( $php_files, $css_files, $other_files ) {
+
+		$php = implode( ' ', $php_files );
+
+		checkcount();
+
+		if ( ! preg_match( '/wp_enqueue_script\(\s?("|\')jquery("|\')/i', $php ) ) {
+			$this->error[] = sprintf(
+				'<span class="tc-lead tc-warning">%s</span>: %s',
+				__( 'RECOMMENDED', 'theme-check' ),
+				sprintf(
+					'JQuery is enqueued on the front end, however you may not need it, see <a href="%s">this article</a> for tips on switching to vanilla JavaScript.',
+					'https://tobiasahlin.com/blog/move-from-jquery-to-vanilla-javascript/'
+				)
+			);
+		}
+		return true;
+	}
+
+	/**
+	 * Get error messages from the checks.
+	 *
+	 * @return array Error message.
+	 */
+	public function getError() {
+		return $this->error;
+	}
+}
+
+$themechecks[] = new JQuery_Check();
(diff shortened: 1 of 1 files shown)


checks/class-jquery-check.php
```diff
@@ -0,0 +1,55 @@
...
+
+		checkcount();
+
+		if ( ! preg_match( '/wp_enqueue_script\(\s?("|\')jquery("|\')/i', $php ) ) {
+			$this->error[] = sprintf(
+				'<span class="tc-lead tc-warning">%s</span>: %s',
+				__( 'RECOMMENDED', 'theme-check' ),
+				sprintf(
+					'JQuery is enqueued on the front end, however you may not need it, see <a href="%s">this article</a> for tips on switching to vanilla JavaScript.',
+					'https://tobiasahlin.com/blog/move-from-jquery-to-vanilla-javascript/'
```
adamsilverstein, line 38: Is linking to an external resource OK here? A little worried that the article could be removed at some future date and this would break. Ideally I would link to something on wordpress.org that explains this.

checks/class-jquery-check.php
```diff
@@ -0,0 +1,55 @@
...
+		}
+		return true;
+	}
+
+	/**
+	 * Get error messages from the checks.
+	 *
+	 * @return array Error message.
+	 */
+	public function getError() {
```
adamsilverstein, line 50: copied this bit from other classes, so i'm assuming it is required?

checks/class-jquery-check.php
aristath: What about What about this one? https://make.wordpress.org/themes/2021/10/04/the-performance-impact-of-using-jquery-in-wordpress-themes/

adamsilverstein: Excellent suggestion @aristath - I'll update the link

adamsilverstein: updated

checks/class-jquery-check.php
```diff
@@ -0,0 +1,55 @@
...
+	 * @param array $css_files File paths and content for CSS files.
+	 * @param array $other_files Folder names, file paths and content for other files.
+	 */
+	public function check( $php_files, $css_files, $other_files ) {
+
+		$php = implode( ' ', $php_files );
+
+		checkcount();
+
+		if ( ! preg_match( '/wp_enqueue_script\(\s?("|\')jquery("|\')/i', $php ) ) {
```
aristath, line 32: This will catch enqueueing jQuery successfully. However, it will not catch cases where jQuery is a dependency of another script - in which case it also gets loaded.
Most themes don't enqueue jQuery directly, it is simply used as a dependency for their own scripts, so we should check for that scenario as well.

aristath (changes requested):

checks/class-jquery-check.php
adamsilverstein: Good call, that's the case that actually matters - very few themes enqueue jQuery by hand. Pushed an update.

_Claude went through the code for this one:_

> The check now looks for two things in each PHP file:
>
> 1. A direct enqueue - `wp_enqueue_script( 'jquery' )`.
> 2. jQuery in the `$deps` array of `wp_enqueue_script()` or `wp_register_script()`, for example `wp_enqueue_script( 'theme-script', $url, array( 'jquery' ), $ver, true )`. Both `array( ... )` and `[ ... ]` are matched, jQuery can sit anywhere in the list, and whitespace or newlines between the arguments don't matter.
>
> Both patterns accept the jQuery handles WordPress registers: `jquery`, `jquery-core`, `jquery-migrate` and the `jquery-ui-*` family. Each of those either is jQuery or depends on it, so any of them being loaded means jQuery is loaded. Handles that merely start with `jquery-` but are the theme's own - `jquery-slider`, say - are not matched.
>
> Two other things changed while in there. The condition was inverted: `if ( ! preg_match( ... ) )` fired the notice when jQuery was *absent*, so it warned every theme that had already dropped jQuery and stayed quiet for the ones it was meant to flag. That came from copying the comment-reply check, where a missing script is the problem. And the notice is now reported per file with `tc_filename()` and `tc_preg()`, like the other checks, so it names the file and shows the matching lines instead of emitting one generic message.
>
> Known limits, both left alone deliberately: a dependency list assigned to a variable first (`$deps = array( 'jquery' ); wp_enqueue_script( 'x', $url, $deps );`) is not detected, and the check does not try to tell a front end enqueue from an admin one, since neither is reachable with a regex over the source.

Also merged trunk in - the branch was four years behind. `phpcs` is clean.

checks/class-jquery-check.php
adamsilverstein: _Answered by Claude:_ both are required - `getError()` is declared on the `themecheck` interface in `checkbase.php`, and the `$themechecks[]` line is the only registration path, since there is no autodiscovery.

Issue #422, opened 2022-04-09
From WP 6.0, block themes do not require index.php


carolinan:
Please see
https://core.trac.wordpress.org/ticket/54272

kafleg:
Thank you @carolinan 
We'll update the plugin accordingly.

PR #424, opened 2022-06-02
Update check for index.php

jffng:
Addresses #422. This PR updates the theme check to check for both `index.php` or `templates/index.html`, since WordPress themes are no longer required to have a index.php file: https://core.trac.wordpress.org/changeset/52940 

_(Started this PR at WCEU Contributor Day, I haven't had a chance to test this yet!)_

### DIFF
--- a/checks/class-file-check.php
+++ b/checks/class-file-check.php
@@ -66,7 +66,11 @@ class File_Check implements themecheck {
 			'favicon\.ico'        => __( 'Favicon', 'theme-check' ),
 		);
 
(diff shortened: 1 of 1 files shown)


carolinan:
Out of the two pull requests for this issue, I prefer this solution, but we also need to check for block-templates/index.html.

acosmin:
@carolinan this one can't possibly work, it will return true if you have index.php in any of the theme folders and the `$filenames` does not include the `templates` path/folder, it will just return index.html.

carolinan:
Got it, that makes perfect sense.

PR #425, opened 2022-06-03
Remove index php requirement

kafleg:
This PR will remove the REQUIRED index.php file from block themes. From WordPress 6.0, index.php file is optional in Block themes.

### DIFF
--- a/checks/class-file-check.php
+++ b/checks/class-file-check.php
@@ -68,13 +68,16 @@ class File_Check implements themecheck {
 
 		$musthave = array( 'index.php', 'style.css', 'readme.txt' );
 
-		$fse_find = array_filter( array_keys( $other_files ), function( $file_name ) {
-			if ( false !== stripos( $file_name, 'templates/index.html' ) || false !== stripos( $file_name, 'block-templates/index.html' ) ) {
-				return true;
-			}
+		$fse_find = array_filter(
+			array_keys( $other_files ),
+			function( $file_name ) {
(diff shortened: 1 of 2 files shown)


acosmin:
This is for #422 

The code works, but it needs to follow the Code Standards

checks/class-file-check.php
```diff
@@ -68,6 +68,19 @@ public function check( $php_files, $css_files, $other_files ) {
 
 		$musthave = array( 'index.php', 'style.css', 'readme.txt' );
 
+		$fse_find = array_filter( array_keys( $other_files ), function( $file_name ) {
+			if ( false !== stripos( $file_name, 'templates/index.html' ) || false !== stripos( $file_name, 'block-templates/index.html' ) ) {
+				return true;
```
carolinan, line 73: Will this not return true for the exception that was brought up on Trac here: https://core.trac.wordpress.org/ticket/54910,
classic themes that has a templates folder with a "silence is golden" index.html file.

acosmin: It probably will :-) but I haven't seen any themes adding it in that way

We could probably add in the mix a check for the `full-site-editing` tag, but it's not future proof...

latracal: [wp_is_block_theme](https://github.com/WordPress/wordpress-develop/blob/0a6c37ceb0a86bebf0b287ac12d200e67bbf948f/src/wp-includes/class-wp-theme.php#L1486) has similar implementation.

acosmin (approved):

Issue #430, opened 2022-09-13
text-domain is not parsed correctly when theme has no i18n translation


latracal:
When theme has no i18n translation, the Theme Check fails with the message - 
`REQUIRED: This theme text domain does not match the theme's slug. The text domain used: This theme's correct slug and text-domain is no-i18n.`

Further, the INFO log for text-domain seems to be rendering an empty text-domain (Notice the 'domain found is .' part) -
`INFO: Only one text-domain is being used in this theme. Make sure it matches the theme's slug correctly so that the theme will be compatible with WordPress.org language packs. The domain found is .`

== Testing Environment ==
PHP Version: 8.1.4
WordPress Version: 6.0.2
Theme Check Version: 20211203 
Theme Slug: no-i18n

== To Replicate ==
I've created a basic theme with no translation to test this out quickly.
Repository - [no-i18n](https://github.com/latracal/no-i18n)
Theme Download Link - [no-i18n.zip](https://github.com/latracal/no-i18n/archive/refs/tags/1.0.0.zip)

justintadlock:
I'll try to dive into this a bit more when I get a chance, but I believe we need to check whether `$domains` is empty here: https://github.com/WordPress/theme-check/blob/master/checks/class-textdomain-check.php#L215

Probably something like:

```php
if ( ! empty( $domains ) && ! in_array( $correct_domain, $domains ) ) {
```

I'm not familiar enough with the internals of TC yet to know if that'll blow something else up yet.

dd32:
Was this resolved by #435?

PR #435, opened 2022-11-17
Checks for empty textdomain array

justintadlock:
This commit addresses issue #430. When a theme doesn't use any of the internationalization functions, the `TextDomain_Check` class doesn't account for an empty array of found domains.


kafleg:
Thank you for this PR @justintadlock I checked and it is working fine. 

Would you please check **PHP Code Standards** because all of them failed in this PR?

PR #436, opened 2022-11-18
Update URI check added

kafleg:
This PR will check if Update URI will be present in the style.css header or not.

If present, it will display a REQUIRED message.


acosmin:
@kafleg this one looks ok too. Maybe in the future, we can add a not allowed array instead of using an `if` clause

checks/class-style-css-header-check.php
```diff
@@ -38,18 +38,31 @@ public function check( $php_files, $css_files, $other_files ) {
...
+                if (preg_match( '/' . $key . '/i', $css, $matches ) ) {
+                    $this->error[] = sprintf(
+                        '<span class="tc-lead tc-required">%s</span> %s',
+                        __( 'REQUIRED', 'theme-check' ),
+                        $check
+                    );
+                    $ret = false;
+                }
+            }else{
+                if ( ! preg_match( '/' . $key . '/i', $css, $matches ) ) {
```
justintadlock, line 57: Does this `if` check need to be nested?  Or, can it be moved up a level so that there's just an `if`/`elseif` check?

justintadlock:
Can we get a refresh of this that uses WP formatting, such as tabs instead of spaces?

checks/class-style-css-header-check.php
acosmin: `elseif` works too.

PR #437, opened 2022-11-21
Checked Full Site Editing Tag in Classic Theme

kafleg:
This PR will check if Classic Theme used full-site-editing tag in style.css header or not.


acosmin:
Looks ok :)

kafleg:
Thanks for checking @acosmin :)

justintadlock:
I added a comment to one bit of the code there about the nested `if` statement.  Also, can we get a refresh of this that uses WP formatting, such as tabs instead of spaces?

checks/class-style-tags-check.php
```diff
@@ -8,227 +8,254 @@
...
+                    $root_dir = get_theme_root($theme_name);
+                    $theme_dir = "$root_dir/$theme_name";
+
+
+                    if (!file_exists($theme_dir . '/theme.json') || (!file_exists($theme_dir . '/templates/index.html') && !file_exists($theme_dir . '/block-templates/index.html'))) {
+                        $this->error[] = sprintf(
+                            '<span class="tc-lead tc-required">%s</span> %s',
+                            __('REQUIRED', 'theme-check'),
+                            sprintf(
+                                __('The tag %s is being  used, please remove it from your style.css header. It is only allowed for Block Theme. ', 'theme-check'),
```
justintadlock, line 107: Removed some extra spaces and some minor grammatical changes:

```php
__( 'The tag %s is being used, please remove it from your style.css header. It is only allowed in block themes.', 'theme-check' ),
```

justintadlock:
The code changes look pretty good to me (see comment above).

Outside of that, it looks like you need to turn on tabs (instead of spaces) for your code editor.  The commit is overwriting the entire file.

Issue #439, opened 2023-03-21
"style-variations" tag is missing from the valid theme tags


matiasbenedetto:
"style-variations" tag is missing from the valid theme tags.
The "style-variations" is valid feature tag documented here:
https://make.wordpress.org/themes/handbook/review/required/theme-tags/#features-tags

PR #440, opened 2023-03-21
adding style-variations tag to the list of valid tags

matiasbenedetto:
## What?
Adding "style-variations" tag is missing from the valid theme tags.

Without this fix you get this warning on themes using that tag:
```
INFO Found wrong tag, remove style-variations from your style.css header.
```

## Why? 
The "style-variations" is valid feature tag documented here:
https://make.wordpress.org/themes/handbook/review/required/theme-tags/#features-tags

Fixes: #439


kafleg:
Thank you for the PR. Looks good to me.

PR #450, opened 2023-04-17
Update Author

kafleg:
Theme review team is now the themes team and it is already changed everywhere. It was missing here.


kafleg:
Merged after review by @acosmin

PR #451, opened 2023-08-09
Add Style Variations tags

kafleg:
Style Variations tag is recently introduced and it was missing in the theme check plugin tag list.

This PR will add tag in the theme check.

Check this PR with TT3 theme.


kafleg:
Merged after reviewed by @acosmin

PR #457, opened 2024-09-03
adds skip links check for block themes

MaggieCabrera:
This PR adds a check only for block themes to see if templates have a main tag present. If they don't they will be missing the skip links from said template.

I need to check when a template is made out of a pattern and the main tag will be inside the pattern instead of the template file

### DIFF
--- a/checks/class-skip-links.php
+++ b/checks/class-skip-links.php
@@ -33,17 +33,7 @@ class Skip_Links_Check implements themecheck {
 	function set_context( $data ) {
 		if ( isset( $data['theme'] ) ) {
 			$this->wp_theme = $data['theme'];
-			$theme_dir      = $this->wp_theme->get_stylesheet_directory();
-			// Check if the theme has all the required files.
-			if (
-				file_exists( $theme_dir . '/theme.json' ) ||
-				(
-					file_exists( $theme_dir . '/templates/index.html' ) &&
-					file_exists( $theme_dir . '/block-templates/index.html' )
-				)
-			) {
-				$this->is_block_theme = true;
-			}
+			$this->is_block_theme = wp_is_block_theme();
 		}
 	}
 
--- a/checks/class-skip-links.php
+++ b/checks/class-skip-links.php
@@ -82,7 +82,7 @@ class Skip_Links_Check implements themecheck {
 				'<span class="tc-lead tc-required">%s</span> %s ',
 				__( 'REQUIRED', 'theme-check' ),
 				sprintf(
-					__( 'Skip links are missing from the following templates: %s Please make sure the templates have a <main> tag', 'theme-check' ),
+					__( 'Skip links are missing from the following templates: %s Please make sure the templates have a &lt;main&gt; tag', 'theme-check' ),
 					$info
 				)
 			);
--- a/checks/class-skip-links.php
+++ b/checks/class-skip-links.php
@@ -49,31 +49,28 @@ class Skip_Links_Check implements themecheck {
 		$info                       = '';
 		$templates_without_main_tag = array();
 
-		$directory = 'templates'; // Path to the folder containing HTML files
-		$theme_dir = $this->wp_theme->get_stylesheet_directory();
+		foreach ( $other_files as $php_key => $file ) {
+			//if the file is a template, print the name of the file
+			if ( strpos( $php_key, 'templates/' ) !== false ) {
 
-		// Get all HTML files in the directory
-		$files = glob( $theme_dir . '/' . $directory . '/*.html' );
+				$file_name = tc_filename( $php_key );
+				$has_main_tag = strpos( $file, '<main' ) !== false;
 
-		foreach ( $files as $file ) {
-			$contents     = file_get_contents( $file );
-			$has_main_tag = strpos( $contents, '<main' ) !== false;
-			$file_name    = basename( $file );
-
-			if ( ! $has_main_tag ) {
-				$pattern_slugs = $this->template_has_patterns( $contents );
-				if ( $pattern_slugs ) {
-					foreach ( $pattern_slugs as $slug ) {
-						$has_main_tag = $this->pattern_has_tag( $slug );
-						if ( ! $has_main_tag ) {
-							if ( ! in_array( $file_name, $templates_without_main_tag ) ) {
-								$templates_without_main_tag[] = $file_name;
+				if ( ! $has_main_tag ) {
+					$pattern_slugs = $this->template_has_patterns( $file );
+					if ( $pattern_slugs ) {
+						foreach ( $pattern_slugs as $slug ) {
+							$has_main_tag = $this->pattern_has_tag( $slug );
+							if ( ! $has_main_tag ) {
+								if ( ! in_array( $file_name, $templates_without_main_tag ) ) {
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
We might want to check if there's multiple tags too

checks/class-skip-links.php
```diff
@@ -0,0 +1,152 @@
...
+			if (
+				file_exists( $theme_dir . '/theme.json' ) ||
+				(
+					file_exists( $theme_dir . '/templates/index.html' ) &&
+					file_exists( $theme_dir . '/block-templates/index.html' )
+				)
+			) {
+				$this->is_block_theme = true;
+			}
+		}
```
matiasbenedetto, line 47: Could we replace this implementation by the use of the core function `wp_is_block_theme()`. Reference:  https://developer.wordpress.org/reference/functions/wp_is_block_theme/

matiasbenedetto (commented):
the `<main>` tag is being rendered in the HTML instead of being printed as text:

image

matiasbenedetto:
It would be nice to make the `main` tag search recusive for nested patterns.

MaggieCabrera:
I fixed those! ready for another review

checks/class-skip-links.php
```diff
@@ -0,0 +1,150 @@
...
+
+		$info = implode( ', ', $templates_without_main_tag );
+
+		if ( $info !== '' ) {
+			$this->error[] = sprintf(
+				'<span class="tc-lead tc-required">%s</span> %s ',
+				__( 'REQUIRED', 'theme-check' ),
+				sprintf(
+					__( 'Skip links are missing from the following templates: %s Please make sure the templates have a &lt;main&gt; tag', 'theme-check' ),
```
matiasbenedetto, line 85:
```suggestion
					__( 'Skip links are missing from the following templates: %s Please make sure the templates have a &lt;main&gt; tag', 'theme-check' ),
```







All the other messages end with a stop.


```suggestion
					__( 'Skip links are missing from the following templates: %s Please make sure the templates have a &lt;main&gt; tag.', 'theme-check' ),
```

checks/class-skip-links.php
```diff
@@ -0,0 +1,150 @@
...
+	public function check( $php_files, $css_files, $other_files ) {
+
+		$info                       = '';
+		$templates_without_main_tag = array();
+
+		$directory = 'templates'; // Path to the folder containing HTML files
+		$theme_dir = $this->wp_theme->get_stylesheet_directory();
+
+		// Get all HTML files in the directory
+		$files = glob( $theme_dir . '/' . $directory . '/*.html' );
```
matiasbenedetto, line 56: Accessing the file system from the `check` function doesn't seems to be necesary and it looks like it's not in line with the other implementation. The `check` function is already getting 3 arrays as parameters containing the list of file paths of the theme.

checks/class-skip-links.php
```diff
@@ -0,0 +1,150 @@
...
+		$theme_dir = $this->wp_theme->get_stylesheet_directory();
+
+		if ( ! is_dir( $theme_dir . '/' . $directory ) ) {
+			$directory = 'block-patterns';
+		}
+		if ( ! is_dir( $theme_dir . '/' . $directory ) ) {
+			return false;
+		}
+
+		$files = glob( $theme_dir . '/' . $directory . '/*.php' );
```
matiasbenedetto, line 115: Same as in https://github.com/WordPress/theme-check/pull/457#discussion_r1761338458

kafleg:
I checked by removing `"tagName":"main"` from the template files in Twenty Twenty Four theme but this PR won't show anything. 

Would you please let me know how did you test that?

kafleg:
> I checked by removing `"tagName":"main"` from the template files in Twenty Twenty Four theme but this PR won't show anything.
> 
> Would you please let me know how did you test that?

I find it. I was checking different PR. Sorry!

kafleg:
REQUIRED Skip links are missing from the following templates: single.html, single.html Please make sure the templates have a <main> tag.

Why the message show `single.html, single.html` twice?

MaggieCabrera:
> REQUIRED Skip links are missing from the following templates: single.html, single.html Please make sure the templates have a
> tag.
> 
> Why the message show `single.html, single.html` twice?

I haven't seen that on my tests, what did you try to get that to happen?

kafleg:
In single.html file of TT4 theme, I simply changed `<main class="wp-block-group alignfull">` to `<div class="wp-block-group alignfull">`

kafleg:
I think I found the reason,
I changed on index.html and single.html in two files and it is happended.

checks/class-skip-links.php
```diff
@@ -0,0 +1,150 @@
...
+		if ( $info !== '' ) {
+			$this->error[] = sprintf(
+				'<span class="tc-lead tc-required">%s</span> %s ',
+				__( 'REQUIRED', 'theme-check' ),
+				sprintf(
+					__( 'Skip links are missing from the following templates: %s Please make sure the templates have a &lt;main&gt; tag.', 'theme-check' ),
+					$info
+				)
+			);
+		}
```
matiasbenedetto, line 89: When a REQUIRED change is return by the `check` function, it should return `false`:
See https://github.com/WordPress/theme-check/issues/465 as reference

MaggieCabrera: Good catch, thanks!

MaggieCabrera:
> I think I found the reason, I changed on index.html and single.html in two files and it is happended.

I just fixed this

checks/class-skip-links.php
MaggieCabrera: It looks like `$php_files` doesn't store the contents of the comment in the pattern, so I have no way of checking if the slug of the pattern is the one I'm trying to match

MaggieCabrera: seems like that's done here so I will leave this part of the code as I implemented it: https://github.com/WordPress/theme-check/blob/b2b4cec4db511691827068efb3d4010911c0d66d/checkbase.php#L61

MaggieCabrera:
Sorry, I have been super busy; I just caught up with the reviews. It should be ready to go now!

checks/class-skip-links.php
```diff
@@ -0,0 +1,152 @@
...
+	 *
+	 * @var WP_Theme $wp_theme
+	 */
+	protected $wp_theme = false;
+
+	function set_context( $data ) {
+		if ( isset( $data['theme'] ) ) {
+			$this->wp_theme       = $data['theme'];
+			$this->is_block_theme = wp_is_block_theme();
+		}
```
matiasbenedetto, line 37: As far as I can see `is_block_theme` is not being used in this check class. Is it used somewhere?

matiasbenedetto (commented):
I think the `Skip_Links_Check` class needs to be added to the list of FSE specific checks here:

https://github.com/WordPress/theme-check/blob/dabaf1d021cd4e44aebbe0deec35a5eda304ec6a/checkbase.php#L386-L387

So this line needs to be removed:

https://github.com/WordPress/theme-check/blob/dabaf1d021cd4e44aebbe0deec35a5eda304ec6a/checks/class-skip-links.php#L152

See `FSE_Required_Files_Check` class as reference.

PR #458, opened 2024-09-06
Add wp-cli integration

matiasbenedetto:
## What?
Add wp-cli integration for theme-check plugin

## Why?
To use theme-coutput of the theme-check plugin in automated envs like CI workflows.

## How to use and how to test
- In your WordPress install `theme-check` plugin **using this branch**.
- Install `wp-cli` in your system.
- Run in your console `wp theme-check run`.
- Observe the output in your console

It should look like this:

Running `wp theme-check run` :
image

Running `wp theme-check run --format=json`:
image

## Screencast
[Screencast from 06-09-24 11:39:55.webm](https://github.com/user-attachments/assets/0add5f81-c3c0-40ee-89f0-ab0e9ea88fde)

### DIFF
--- a/wp-cli/class-theme-check-cli.php
+++ b/wp-cli/class-theme-check-cli.php
@@ -1,0 +1,176 @@
+<?php
+
+if ( defined( 'WP_CLI' ) && WP_CLI ) {
+    $parent_dir = dirname(__DIR__);
+    require_once $parent_dir . '/checkbase.php';
+    require_once $parent_dir . '/main.php';
+    WP_CLI::add_command( 'theme-check', 'Theme_Check_CLI' );
+}
+
+class Theme_Check_CLI {
+    /**
+     * Run a theme check on the specified theme or the current theme.
+     *
+     * ## OPTIONS
+     *
+     * [<theme>]
+     * : The slug of the theme to check. If not provided, checks the current theme.
+     *
+     * [--format=<format>]
+     * : Output format. Accepts 'cli' or 'json'. Default: 'cli'.
+     *
+     * ## EXAMPLES
+     *
+     *     wp theme-check run
+     *     wp theme-check run twentytwentyfour
+     *     wp theme-check run --format=json
+     *     wp theme-check run twentytwentyfour --format=json
+     *
+     */
+    public function run( $args, $assoc_args ) {
+        // Get the output format
+        $format = \WP_CLI\Utils\get_flag_value( $assoc_args, 'format', 'cli' );
+        if ( !in_array( $format, ['cli', 'json'] ) ) {
+            WP_CLI::error( "Invalid format. Accepts 'cli' or 'json'." );
+            return;
+        }
+
+        // Get the current theme
+        $current_theme = wp_get_theme();
+        $current_theme_slug = $current_theme->get_stylesheet();
+
+        // Use the provided theme slug if available, otherwise use the current theme
+        $check_theme_slug = $args[0] ?? $current_theme_slug;
+
+        // Get the theme
+        $theme = wp_get_theme( $check_theme_slug );
+
+        if ( ! $theme->exists() ) {
+            if ( $format === 'json' ) {
+                $json_output = array(
+                    'check-completed' => false,
+                    'result' => "Error: Theme '{$check_theme_slug}' not found.",
+                    'messages' => []
+                );
+                WP_CLI::log (wp_json_encode($json_output, JSON_PRETTY_PRINT) );
+                return;
+            }
+            WP_CLI::error( "Theme '{$check_theme_slug}' not found." );
+            return;
+        }
+
+        // Run the checks.
+        $success = run_themechecks_against_theme( $theme, $check_theme_slug );
+
+        if ( $format === 'json' ) {
+            if ( ! $success ) {
+                $json_output = array(
+                    'check-completed' => false,
+                    'result' => "Error: Theme check failed for {$slug}.",
+                    'messages' => []
+                );
+                WP_CLI::log (wp_json_encode($json_output, JSON_PRETTY_PRINT) );
+                return;
+            }
+            $this->display_themechecks_as_json( $check_theme_slug );
+        } else {
+            if ( ! $success ) {
+                WP_CLI::error( "Theme check failed for {$slug}." );
+                return;
+            }
+            $this->display_themechecks_in_cli( $check_theme_slug );
+        }
+    }
+
+    /**
+     * Process theme check messages.
+     *
+     * @return array Processed messages categorized by type.
+     */
+    private function process_themecheck_messages() {
+        global $themechecks;
+        $messages = array();
+        $processed_messages = array(
+            'errors' => array(),
+            'warnings' => array(),
+            'infos' => array(),
+            'others' => array()
+        );
+
+        foreach ($themechecks as $check) {
+            if ($check instanceof themecheck) {
+                $error = $check->getError();
+                $error = (array) $error;
+                if (!empty($error)) {
+                    $messages = array_merge($messages, $error);
+                }
+            }
+        }
+
+        foreach ($messages as $message) {
+            $clean_message = html_entity_decode(strip_tags($message), ENT_QUOTES, 'UTF-8');
+
+            if (strpos($clean_message, 'ERROR:') === 0) {
+                $processed_messages['errors'][] = $clean_message;
+            } elseif (strpos($clean_message, 'WARNING:') === 0) {
+                $processed_messages['warnings'][] = $clean_message;
+            } elseif (strpos($clean_message, 'INFO') === 0) {
+                $processed_messages['infos'][] = $clean_message;
+            } else {
+                $processed_messages['others'][] = $clean_message;
+            }
+        }
+
+        return $processed_messages;
+    }
+
+    /**
+     * Display the theme checks in the CLI.
+     *
+     * @return void
+     */
+    private function display_themechecks_in_cli( $slug ) {
+        $processed_messages = $this->process_themecheck_messages();
+
+        WP_CLI::success("Theme check completed for {$slug}.");
+
+        if (
+            !empty($processed_messages['errors']) ||
+            !empty($processed_messages['warnings']) ||
+            !empty($processed_messages['infos']) ||
+            !empty($processed_messages['others'])
+        ) {
+            foreach ($processed_messages['errors'] as $error) {
+                WP_CLI::error(ltrim($error, 'ERROR: '), false);
+            }
+            foreach ($processed_messages['warnings'] as $warning) {
+                WP_CLI::warning(ltrim($warning, 'WARNING: '));
+            }
+            foreach ($processed_messages['infos'] as $info) {
+                WP_CLI::log( $info );
+            }
+            foreach ($processed_messages['others'] as $other) {
+                WP_CLI::log("LOG: " . $other);
+            }
+        } else {
+            WP_CLI::success("No issues found.");
+        }
+    }
+
+    /**
+     * Display the theme checks in JSON format.
+     *
+     * @return void
+     */
+    private function display_themechecks_as_json( $slug ) {
+        $processed_messages = $this->process_themecheck_messages();
+
+        $json_output = array(
+            'check-completed' => true,
+            'result' => "Theme check completed for {$slug}.",
+            'messages' => $processed_messages
+        );
+
+        WP_CLI::log(wp_json_encode($json_output, JSON_PRETTY_PRINT));
+    }
+}
--- a/theme-check.php
+++ b/theme-check.php
@@ -14,8 +14,12 @@ class ThemeCheckMain {
 	function __construct() {
 		add_action( 'admin_init', array( $this, 'tc_i18n' ) );
 		add_action( 'admin_menu', array( $this, 'themecheck_add_page' ) );
-	}
 
+		if ( defined( 'WP_CLI' ) && WP_CLI ) {
+			include 'wp-cli/class-theme-check-cli.php';
+		}
+	}
+	
 	function tc_i18n() {
 		load_plugin_textdomain( 'theme-check', false, dirname( plugin_basename( __FILE__ ) ) . '/lang/' );
 	}
@@ -43,6 +47,7 @@ class ThemeCheckMain {
 
 		include 'checkbase.php';
 		include 'main.php';
+		
 
 		?>
 		<div id="theme-check" class="wrap">


wp-cli/class-theme-check-cli.php
```diff
@@ -0,0 +1,179 @@
...
+					'result'          => "Error: Theme check failed for {$slug}.",
+					'messages'        => array(),
+				);
+				WP_CLI::log( wp_json_encode( $json_output, JSON_PRETTY_PRINT ) );
+				return;
+			}
+			$this->display_themechecks_as_json( $check_theme_slug );
+		} else {
+			if ( ! $success ) {
+				WP_CLI::error( "Theme check failed for {$slug}." );
```
jffng, line 79:
```suggestion
				WP_CLI::error( "Theme check failed for {$check_theme_slug}." );
```

wp-cli/class-theme-check-cli.php
```diff
@@ -0,0 +1,179 @@
...
+		}
+
+		// Run the checks.
+		$success = run_themechecks_against_theme( $theme, $check_theme_slug );
+
+		if ( $format === 'json' ) {
+			if ( ! $success ) {
+				$json_output = array(
+					'check-completed' => false,
+					'result'          => "Error: Theme check failed for {$slug}.",
```
jffng, line 70:
```suggestion
					'result'          => "Error: Theme check failed for {$check_theme_slug}.",
```

jffng (commented):
When the theme check fails, the error messages are not logged. All that is returned is: 

```
$ wp theme-check run
Error: Theme check failed for twentytwentyfive.
```

jffng:
Nice change! I can't push to your branch, but here is some suggestion changes to the readme:


### Usage with wp-cli

To use with [wp-cli](https://wp-cli.org/), ensure the theme check plugin is active and `wp-cli` is installed. The `theme-check` subcommand is added to `wp-cli` and can be used as follows:

`wp theme-check run [<theme>] [--format=<format>]`

#### Options
| Option | Accepts | Required | Default
| -- | -- | -- | -- | 
| `theme` | The slug of the theme to check | No | Current theme slug
| `format` | `cli` or `json` | No | `cli`

#### Examples
`wp theme-check run`
`wp theme-check run twentytwentyfour`
`wp theme-check run --format=json`
`wp theme-check run twentytwentyfour --format=json`

matiasbenedetto:
@jffng cool, look good. I added you as a collaborator in the fork so you can push :)

codestylist:
What a cool feature @matiasbenedetto . Thank you for this! Hope it will be in the master soon!

wp-cli/class-theme-check-cli.php
```diff
@@ -0,0 +1,179 @@
...
+	 *
+	 * @param array $args       Indexed array of positional arguments.
+	 * @param array $assoc_args Associative array of options.
+	 * @return void
+	 *
+	 * ## OPTIONS
+	 * [<theme>]
+	 * : The slug of the theme to check. If not provided, checks the current theme.
+	 *
+	 * [--format=<format>]
```
ernilambar, line 22: If you follow this format https://github.com/wp-cli/entity-command/blob/main/src/Post_Command.php#L419 then WP-CLI itself will validate the available options. That way you won't need `in_array()` check in the method.

wp-cli/class-theme-check-cli.php
```diff
@@ -0,0 +1,179 @@
...
+				$json_output = array(
+					'check-completed' => false,
+					'result'          => "Error: Theme '{$check_theme_slug}' not found.",
+					'messages'        => array(),
+				);
+				WP_CLI::log( wp_json_encode( $json_output, JSON_PRETTY_PRINT ) );
+				return;
+			}
+			WP_CLI::error( "Theme '{$check_theme_slug}' not found." );
+			return;
```
ernilambar, line 60: WP_CLI::error() exits the process. So I don't think `return` is needed.

wp-cli/class-theme-check-cli.php
```diff
@@ -0,0 +1,179 @@
...
+		$success = run_themechecks_against_theme( $theme, $check_theme_slug );
+
+		if ( $format === 'json' ) {
+			if ( ! $success ) {
+				$json_output = array(
+					'check-completed' => false,
+					'result'          => "Error: Theme check failed for {$check_theme_slug}.",
+					'messages'        => array(),
+				);
+				WP_CLI::log( wp_json_encode( $json_output, JSON_PRETTY_PRINT ) );
```
ernilambar, line 73: Note that output of `log()` is discarded to STDOUT when `--quiet` flag is passed. But here I think we should display JSON output whatever case if quiet flag is present or not.

wp-cli/class-theme-check-cli.php
```diff
@@ -0,0 +1,179 @@
...
+	 * @return void
+	 *
+	 * ## OPTIONS
+	 * [<theme>]
+	 * : The slug of the theme to check. If not provided, checks the current theme.
+	 *
+	 * [--format=<format>]
+	 * : Output format. Accepts 'cli' or 'json'. Default: 'cli'.
+	 *
+	 * ## EXAMPLES
```
ernilambar, line 25: Recommended to follow this format with short explanation of the command. https://github.com/wp-cli/entity-command/blob/main/src/Post_Command.php#L467

wp-cli/class-theme-check-cli.php
matiasbenedetto: If I removed that snippet:
```
		if ( ! in_array( $format, array( 'cli', 'json' ) ) ) {
			WP_CLI::error( "Invalid format. Accepts 'cli' or 'json'." );
			return;
		}
```
If I use a non-existent format the function runs using the default format instead of raising an error. No matter the comment format implemented in the function. Is that expected?

wp-cli/class-theme-check-cli.php
matiasbenedetto: done here: https://github.com/WordPress/theme-check/pull/458/commits/75c2f9cf123ebb6ec2272436cf421429dd6ad7d8

wp-cli/class-theme-check-cli.php
matiasbenedetto: It's not working like that on my end. Removing that snippet and using a random format doesn't throw any error:

image

wp-cli/class-theme-check-cli.php
matiasbenedetto: Ok, I used `line()` to log the results. 

I found 2 issues regarding wp-cli loggers while working on this:
- https://github.com/wp-cli/wp-cli/pull/5979
- https://github.com/wp-cli/wp-cli/issues/5980

wp-cli/class-theme-check-cli.php
matiasbenedetto: :+1: removed here: https://github.com/WordPress/theme-check/pull/458/commits/37542b38dd2ac7b8d2125f67dcc4cfc59dfa4541

wp-cli/class-theme-check-cli.php
```diff
@@ -0,0 +1,157 @@
...
+	require_once $parent_dir . '/checkbase.php';
+	require_once $parent_dir . '/main.php';
+	WP_CLI::add_command( 'theme-check', 'Theme_Check_Command' );
+}
+
+class Theme_Check_Command extends WP_CLI_Command {
+	/**
+	 * Run a theme check on the specified theme or the current theme.
+	 *
+	 * @param array $args       Indexed array of positional arguments.
```
ernilambar, line 14: These 3 lines are breaking the expected doc format of WP-CLI command. So format validation is not working. You can confirm it by checking `wp help theme-check run`. Currently you wont see arguments properly. When you remove these lines, it will work as expected.

ernilambar: You can move these after Examples.

matiasbenedetto: Good. Done here: https://github.com/WordPress/theme-check/pull/458/commits/e998dc02035639c6f0bb3fa297842a09391abd95

wp-cli/class-theme-check-cli.php
matiasbenedetto: It's working fine after implementing your comment formatting suggestion (https://github.com/WordPress/theme-check/pull/458#discussion_r1750604286), thanks!

.vscode/launch.json
```diff
@@ -0,0 +1,48 @@
+{
+    // Use IntelliSense to learn about possible attributes.
```
ernilambar, line 2: May be we need to update gitignore to ignore this type of IDE files and folders.

ernilambar:
When I run check on a theme with errors. `wp theme-check run mazebox`. Output is:

```
Error: Theme check failed for mazebox.
```

But when there is no errors, output includes recommendation and info items.

```
Success: Theme check completed for mazebox.

RECOMMENDED: No reference to register_block_pattern was found in the theme. Theme authors are encouraged to implement custom block patterns as a transition to block themes.

RECOMMENDED: No reference to register_block_style was found in the theme. Theme authors are encouraged to implement new block styles as a transition to block themes.
...

```

.vscode/launch.json
matiasbenedetto: I commited that file accidentally. I removed it here: https://github.com/WordPress/theme-check/pull/458/commits/8d61a5995bed77cdcf42dad6259e7549bb0f61f4

I will update the .gitignore file in another PR.

matiasbenedetto: Submitted a PR for the gitignore: https://github.com/WordPress/theme-check/pull/460

matiasbenedetto:
> When I run check on a theme with errors. wp theme-check run mazebox. Output is:
> Error: Theme check failed for mazebox.

I don't know what kind of errors are you checking but I think this is expected. Theme-check will return an error message if the check couldn't be run for some reason (errors parsing theme code for example). The success and error messages are only used to signal if the plugin was able to run the checks or not, not to output the results of those checks.

> But when there is no errors, output includes recommendation and info items.

It's a little bit counterintuitive, but the theme-check plugin doesn't work with the objective of outputting a boolean ("theme OK/ theme with errors"). The plugin is not designed like that. It just outputs a list of messages regarding the theme, but it never says whether this theme is OK or not. Maybe in the future, we could implement that, but that's not the case yet.

wp-cli/class-theme-check-cli.php
vcanales: > But here I think we should display JSON output whatever case if quiet flag is present or not.

Not sure I agree with bypassing `--quiet`, I would expect no output at all. An alternative is to make sure that the program always exits with a descriptive code; ie. 0 for OK, non-0 for errors.

edit: On reading @matiasbenedetto's comment below, seems like the checks are not meant to return a boolean.

> It's a little bit counterintuitive, but the theme-check plugin doesn't work with the objective of outputting a boolean ("theme OK/ theme with errors"). The plugin is not designed like that. It just outputs a list of messages regarding the theme, but it never says whether this theme is OK or not. Maybe in the future, we could implement that, but that's not the case yet.

Still, I don't think bypassing `--quiet` is cool.

ernilambar:
Example:

```
$ wp theme install twentysixteen --activate
Installing Twenty Sixteen (3.3)
Downloading installation package from https://downloads.wordpress.org/theme/twentysixteen.3.3.zip...
Using cached file '/Users/nilambarsharma/.wp-cli/cache/theme/twentysixteen-3.3.zip'...
Unpacking the package...
Installing the theme...
Theme installed successfully.
Activating 'twentysixteen'...
Success: Switched to 'Twenty Sixteen' theme.
Success: Installed 1 of 1 themes.

$ wp theme install twentyten --activate --quiet
$ echo $?
0
```

Another case:

```
$ wp theme list --format=json --quiet
[{"name":"twentytwelve","status":"inactive","update":"none","version":"4.3","update_version":"","auto_update":"off"}]

$ wp theme list --format=json
[{"name":"twentytwelve","status":"inactive","update":"none","version":"4.3","update_version":"","auto_update":"off"}]
```

There is no right or wrong approach. We can go for any approach we agree upon. But from my experience, general expectation in CLI is that if I pass `--format=json` then I would expect output in valid JSON format, whether quiet flag is passed or not.

vcanales:
> There is no right or wrong approach. We can go for any approach we agree upon. But from my experience, general expectation in CLI is that if I pass `--format=json` then I would expect output in valid JSON format, whether quiet flag is passed or not.

I see now. Yes, I agree that if we're explicitly asking for a format, it makes sense to bypass `--quiet`. On the other hand, these seem like incompatible flags to me; why would someone use both? This is an unrelated discussion, though.

wp-cli/class-theme-check-cli.php
```diff
@@ -0,0 +1,160 @@
...
+				);
+				WP_CLI::line( wp_json_encode( $json_output, JSON_PRETTY_PRINT ) );
+				return;
+			}
+			$this->display_themechecks_as_json( $check_theme_slug );
+		} else {
+			if ( ! $success ) {
+				WP_CLI::error( "Theme check failed for {$check_theme_slug}." );
+			}
+			$this->display_themechecks_in_cli( $check_theme_slug );
```
jffng, line 88: Execution stops if a `WP_CLI::error` is thrown, so the errors are not being displayed. Could this line be moved above the $success check?

matiasbenedetto: I'm not sure, i'm seeing this logic here:
```php
// If the check process failed (if the checks were not able to run properly) raise an error. There's nothing else to do.
if ( ! $success ) {
WP_CLI::error( "Theme check failed for {$check_theme_slug}." );
}
// If the checking process run print in the console the checks. This will also print the things marked as an error in the theme.
$this->display_themechecks_in_cli( $check_theme_slug );
```
Could you explain why you think we should move it?

jffng:
```
// If the check process failed (if the checks were not able to run properly) raise an error. There's nothing else to do.
```

I don't think this is correct. `$success` is the result of `run_themechecks_against_theme` which returns `false` if the theme fails a check. 

> Could you explain why you think we should move it?

Using `twentytwentyfive`, here is the output currently: 

```
Error: Theme check failed for twentytwentyfive.
```

Moving call to `display_themechecks_in_cli` above the conditional, the result is:

```
Success: Theme check completed for twentytwentyfive.

REQUIRED: phpcs.xml.dist XML file found. This file must not be in the production version of the theme.

WARNING: assets/images/image-from-rawpixel-id-2222755.webp is 997.6 KB in size. Large file sizes have a negative impact on website performance and loading time. Compress images before using them.

WARNING: assets/images/image-from-rawpixel-id-2224378.webp is 879.6 KB in size. Large file sizes have a negative impact on website performance and loading time. Compress images before using them.

WARNING: assets/images/location.webp is 778.1 KB in size. Large file sizes have a negative impact on website performance and loading time. Compress images before using them.

WARNING: assets/images/poster-image-background.webp is 574.7 KB in size. Large file sizes have a negative impact on website performance and loading time. Compress images before using them.

INFO Themes that use the tag accessibility-ready will need to undergo an accessibility review. See https://make.wordpress.org/themes/handbook/review/accessibility/

WARNING: Your theme appears to be in the wrong directory for the theme name. The directory name must match the slug of the theme. This theme's correct slug and text-domain is twenty-twenty-five. (If this is a child theme, you can ignore this error.)

INFO: Only one text-domain is being used in this theme. Make sure it matches the theme's slug correctly so that the theme will be compatible with WordPress.org language packs. The domain found is twentytwentyfive.

RECOMMENDED: Requires PHP is recommended to have major and minor versions only (e.g. 7.4). Patch version is not needed (e.g. 7.4.1).

Error: Theme check failed for twentytwentyfive.
```

I think the latter is more useful because it reports what checks failed (the first one marked `REQUIRED`). 

Also I am not sure we should throw an error here if `$success` isn't true, because the theme check successfully ran, but the result was it failed the checks.

ernilambar: This command's focus should be on pass/fail check for a theme, rather than run is successfully completed or not. And exit code should be returned based on that I believe.

ernilambar:
Just an idea: What if we simply list check results like CLI displays. Below is one example. 
Each message is splitted into `type` and `message`. This way we could support different format using WP_CLI\Formatter(), like `table`, `json`, `csv`, `yaml`.

When I need to check in my CI runner where I just need to check `required` items or not, I could use:

```
wp theme-check run theme-slug --type=error --format=json
```

Screenshot 2024-09-11 at 11 18 34 PM

matiasbenedetto:
I went through the theme-check plugin code again, and I found that we can get a boolean signaling if the theme has required changes or not. Knowing that I was able to I implemented all the latest feedback on the last commit:

- Exit code depending on the existence of required changes or not. https://github.com/matiasbenedetto/theme-check/blob/d82740ea74277dd3e776019594ab10923cfbdd60/wp-cli/class-theme-check-cli.php#L86
- Simpler code
- Use the wp-cli table formatting by default. 

These are the results using TT4 theme (I removed `theme.json` file and screenshot image to get REQUIRED changes):

```
wp theme-check run
```
image

---
```
wp theme-check run --format=json
```
image


---
```
wp theme-check run non-existing-theme
```
image

wp-cli/class-theme-check-cli.php
matiasbenedetto: I think I could implement your feedback, please see: https://github.com/WordPress/theme-check/pull/458#issuecomment-2349033538

README.md
```diff
@@ -50,5 +50,23 @@ add_filter( 'tc_skip_development_directories', '__return_true' );
 
 To add more directories to the paths where other files are excluded then add them to the array through the `tc_common_dev_directories` filter.
 
+### Usage with wp-cli
+
+To use with [wp-cli](https://wp-cli.org/), ensure the theme check plugin is active and `wp-cli` is installed. The `theme-check` subcommand is added to `wp-cli` and can be used as follows:
+
+`wp theme-check run [<theme>] [--format=<format>]`
+
```
jffng, line 58:
```suggestion

On success, the command returns a formatted table of results from the theme check plugin.

```

jffng (approved):
It works well in my testing.

One question that does not have to be a blocker, but do you know how to add the help text for the `theme-check` subcommand when a user runs `wp help`?

Screenshot 2024-09-16 at 4 35 38 PM

matiasbenedetto:
> One question that does not have to be a blocker, but do you know how to add the help text for the theme-check subcommand when a user runs wp help?

Added it here: https://github.com/WordPress/theme-check/pull/458/commits/d60d1e3c4d563d06d3d5a4fe721d05afd23ac6ed

Screenshot from 2024-09-16 17-57-05

wp-cli/class-theme-check-cli.php
matiasbenedetto: I think this was fixed on the latest commits.

Issue #465, opened 2024-09-17
Non_GPL_Check check should return false when it fails


matiasbenedetto:
## What?
Non_GPL_Check check should return false when it fails because the error message is a REQUIRED change.

https://github.com/WordPress/theme-check/blob/22226a08d9e13c6b887b72d51e6830bee0b18a1a/checks/class-non-gpl-check.php#L26-L65

## Why?
The REQUIRED changes should return false in the check methods.
Example 1:
https://github.com/WordPress/theme-check/blob/22226a08d9e13c6b887b72d51e6830bee0b18a1a/checks/class-bad-things-check.php#L57-L68

Example 2:
https://github.com/WordPress/theme-check/blob/22226a08d9e13c6b887b72d51e6830bee0b18a1a/checks/class-basic-check.php#L58-L67

PR #466, opened 2024-09-17
Non_GPL_Check check returns false when it fails

matiasbenedetto:
## What?

Non_GPL_Check check returns false when it fails

## Why?
Fixes: https://github.com/WordPress/theme-check/issues/465


vcanales (approved):
~> Non_GPL_Check check returns false when it fails~

~Did you mean that it _should_ return false? If so, this looks good to me.~

LGTM, I read this wrong :)

PR #472, opened 2024-10-21
added comments for translators

mehrazmorshed:
added comments for translators in multiple files

### DIFF
--- a/checks/class-comments-check.php
+++ b/checks/class-comments-check.php
@@ -43,7 +43,7 @@ class Comments_Check implements themecheck {
 					'<span class="tc-lead tc-recommended">%s</span>: %s',
 					__( 'RECOMMENDED', 'theme-check' ),
 					sprintf(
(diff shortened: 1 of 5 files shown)


matiasbenedetto (changes requested):
I haven't found a specific docs about this but for consistency with this plugin codebase and with WordPress core I think the comments should have the  multiline format (`/* translators: Hi! */`)  and not the single line one ( `// translators: Hi! ` ).

Examples from this repo:
https://github.com/search?q=repo%3AWordPress%2Ftheme-check%20translators&type=code

Examples from core:
https://github.com/search?q=repo%3AWordPress%2Fwordpress-develop%20translators&type=code

mehrazmorshed:
Hi, @matiasbenedetto, 

Thank you for the feedback and for pointing out the examples for consistency. 

I have updated the comments to use the multiline format (/* translators: */) as suggested. 

Please let me know if there’s anything else I should address!

matiasbenedetto (approved):
LGTM :+1:

Issue #478, opened 2025-02-11
Update textdomain exceptions to include newer default themes.


joedolson:
The list of [textdomain exceptions](https://github.com/WordPress/theme-check/blob/b2b4cec4db511691827068efb3d4010911c0d66d/checks/class-textdomain-check.php#L61) for default themes hasn't been updated for a while, so it throws warnings on default themes. 

See: https://wordpress.slack.com/archives/C02RQBWTW/p1739294823518509

cc @desrosj

PR #479, opened 2025-02-11
Update class-textdomain-check.php

joedolson:
Add textdomain exceptions for default themes Twenty Twenty-Two, Twenty Twenty-Three, Twenty Twenty-Four, and Twenty Twenty-Five,


joedolson:
See #478

desrosj (approved):

PR #482, opened 2025-09-09
Title Check: Avoid Deprecated warnings when there's no file contents.

dd32:
Not 100% sure how this is happening, but seems like a quick win to avoid unnecessary processing and break early.

```
Deprecated: strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated in wp-content/plugins/theme-check/checks/class-title-check.php on line 57 
```

### DIFF
--- a/checks/class-title-check.php
+++ b/checks/class-title-check.php
@@ -31,7 +31,7 @@ class Title_Check implements themecheck {
 
 		foreach ( $php_files as $file_path => $file_content ) {
 
-			if ( empty( $file_content ) ) {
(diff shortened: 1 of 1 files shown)


kafleg (approved):
Looks good to me.

carolinan:
Note that with the SVG title being fixed in https://github.com/WordPress/theme-check/commit/dfbfafd1462ae9da73001e69644a9c4903dc67b2  Astra cannot be used to test this anymore, the PHP warning has already been fixed. We should still fix this, we will need another theme to test on.

dd32:
Merged #477 closing this.

PR #489, opened 2026-03-19
Skip classic theme checks when the block theme is in a subfolder

### LINKED ISSUE #318: tc_adapt_checks_for_fse_themes fails if the block theme is in a sub folder
When a block based theme is placed inside a subfolder of /`themes`/,  the check for `block-templates/index.html`
inside `tc_adapt_checks_for_fse_themes` fails and checks for classic themes are used instead.

carolinan:
Closes https://github.com/WordPress/theme-check/issues/318

Testing instructions:
Move a block theme to a subfolder of the themes folder.
Run Theme Check on that theme.
Confirm that the classic checks like post thumbnail etc. are not showing in the result.
(Ssee the list of checks that are expected to be skipped: https://github.com/WordPress/theme-check/blob/master/checkbase.php#L359)

### DIFF
--- a/checkbase.php
+++ b/checkbase.php
@@ -351,7 +351,16 @@ function tc_adapt_checks_for_fse_themes( $php_files, $css_files, $other_files ) 
 	}
 
 	// Check whether this is a FSE theme by searching for an index.html block template.
-	if ( ! in_array( 'block-templates/index.html', $other_filenames, true ) && ! in_array( 'templates/index.html', $other_filenames, true ) ) {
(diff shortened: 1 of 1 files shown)


kafleg:
Hi @carolinan 

I checked this PR, by adding a block theme inside a sub folder of Astra theme and test it. Here is the result,
Screenshot 2026-05-18 at 08 49 00

I changed the screenshot to something else, changed the tag and the tag error is showing as info. 

Thank you

carolinan:
@kafleg I think there is some confusion. This has nothing to do with **child theme** checks:

Move a block theme to **a subfolder of the themes folder**.

PR #496, opened 2026-07-20
Updated WordPress 7.0

murshed:
* Fix: Ensure standard translation checks (I18N) are not bypassed when Polylang functions are present in the same file.
* Fix: Target the theme's main style.css file for header checks rather than parsing combined CSS text.
* Update: Verified compatibility with PHP 8.5+.
* Tested with WordPress 7.0.2


kafleg:
@murshed Can you explain how to check this PR? Please explain step by step so that we can review this PR before merging.

